PR #4, opened 2023-10-02
Removing CompositeKEMParams

ounsworth:


johngray-dev (approved):
Looks good!   Other than needing to define the ASN.1 module OID:
id-mod-composite-kems

I was able to compile the ASN.1, so that is great!

ounsworth:
> Looks good! Other than needing to define the ASN.1 module OID: id-mod-composite-kems
> 
> I was able to compile the ASN.1, so that is great!

Yeah, I replace the `TBDMOD` with `999` when I'm test-compiling, but I've gotta leave something that's easy for IANA to CTRL+F replace.

PR #6, opened 2023-10-04
Demoted RSA 3072 to 128-bit Level 1

ounsworth:
This relates to the LAMPS mailist discussion:

https://mailarchive.ietf.org/arch/msg/spasm/xSKIRjtMSJ8M7N7Fx3X09ASUHC8/

The main branch is option 3.1.
This PR is option 3.2.

We should wait to merge this PR pending the results of that discussion thread.

### DIFF
--- a/Composite-KEM-2023.asn
+++ b/Composite-KEM-2023.asn
@@ -102,22 +102,6 @@ kema-CompositeKEM {
 -- Composite KEM Algorithms
 --
 
--- TODO: OID to be replaced by IANA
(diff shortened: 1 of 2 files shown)


Composite-KEM-2023.asn
johngray-dev, line 1: On Line 12 you do not need to import SIGNATURE-ALGORITHM or ParamOptions.   I removed them and was able to compile this ASN.1

johngray-dev (changes requested):

johngray-dev:
This was already included in the john_integration branch Merge #8

PR #11, opened 2024-01-27
Aligned with X-Wing

### LINKED ISSUE #2: WGLC issue: Remove Informative reference to draft-cfrg-kem-combiners ?
This is a procedural issue: this draft is adopted by LAMPS, but it depends Informatively on draft-ounsworth-cfrg-kem-combiners, which is not yet adopted by CFRG. That will cause a problem at WGLC.

I propose to leave this issue for now and revisit it as we approach WGLC.

### LINKED ISSUE #26: Compliant KEM Combiner
In NIST's guidelines for key derivation (see https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-56Cr2.pdf) in section 2 they specify an approved method for combining a compliant secret with auxiliary shared secret like so:

> In addition to the currently approved techniques for the generation of the shared secret Z as
> specified in SP 800-56A and SP 800-56B, this Recommendation permits the use of a “hybrid”
> shared secret of the form Z′ = Z || T, a concatenation consisting of a “standard” shared secret Z that
> was generated during the execution of a key-establishment scheme (as currently specified in [SP
> 800-56A] or [SP 800-56B]) followed by an auxiliary shared secret T that has been generated using
> some other method. The content, format, length, and method used to generate T must be known
> and agreed upon by all parties that will rely upon the derived keying material, as well as by any
> agents trusted to act on their behalf.

The present combiner in this draft doesn't include the secret from an approved method first in the input of the KDF:

> Combiner(ct1, ss1, ct2, ss2, fixedInfo) =
>   KDF(counter || ct1 || ss1 || ct2 || ss2 || fixedInfo, outputBits)

We suggest moving either ss1 or ss2 to the beginning of the input to the KDF since concatenating any auxiliary secret to a compliant secret will result in a compliant key when feeding it to a compliant KDF. Even better, if we choose the classical key to be at the beginning (ss2 in the proposed combinations) it would mean that this construction would be considered compliant by NIST even if used with a non-standardized/compliant post-quantum algorithm.

The fixed counter value was added to in this draft to be compliant, but in the nist specification the counter must be followed by the compliant secret. In this specification ct1 is not a compliant secret, so this doesn't seem to be a good solution.

### LINKED ISSUE #27: Place AlgID at the end of FixedInfo
Falko said:
> In order to avoid any possible collisions in the input it is important 
the algID is placed at the end of the fixedInfo string. This ensures 
that the string is uniquely parsable from the rear end, because algID 
defines the lengths of the other strings.

https://mailarchive.ietf.org/arch/msg/spasm/3Rul4GQ6kgpVZUXaE-kbr-bV9Ck/

Needs to be done after #20 .

ounsworth:


draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -133,50 +132,56 @@ informative:
...
+      -
+        ins: P. Schwabe
+        name: Peter Schwabe
+      -
+        ins: K. Varner
+        name: Karolin Varner
+      -
+        ins: B. Westerbaan
+        name: Bas Westerbaan
+    target: https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.203.ipd.pdf
```
bwesterb: Nice to see X-Wing is a FIPS standard ;).

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -670,7 +711,11 @@ This document uses directly the KEM Combiner defined in {{I-D.ounsworth-cfrg-kem
...
+The component KEMs used in this specification are RSA-KEM {{I-D.ietf-lamps-rfc5990bis}}, ECDH, X25519, X448, and ML-KEM {{FIPS.203-ipd}} all of which meet the criteria of having constant-length shared secrets and ciphertexts and therefore we justify using the simpler construction that omits the length tag.
+
+### Ciphertext collision resistance {#sec-cons-ct-collision}
+
+The notion of a ciphertext cellision resistant KEM is defined in [X-Wing] being the property that it is computationally difficult to find two different ciphertexts that will decapsulate to the same shared secret under the same public key. In [X-Wing] it is proven that ML-KEM has this property and therefore the ML-KEM ciphertext can safely be omitted from the KEM combiner. Similarly, RSA-KEM {{I-D.ietf-lamps-rfc5990bis}} is a bijective map between the shared secret `Z` and the ciphertext `C` and so is naturally immune to ciphertext collision. This leaves only ECDH, X25519 and X448 which are not ciphertext collision resistant and for which the ciphertexts need to be included in the derivation of the shared secret.
```
dconnolly: 👍

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -464,6 +471,11 @@ Each registered composite KEM algorithm must specify the choice of `KDF`, `fixed
 
 See {{sec-cons-kem-combiner}} for further discussion of the security coniserations of this KEM combiner.
 
+However, optimazations may be made in the following ways:
```
dconnolly:
```suggestion
However, optimizations may be made in the following ways:
```

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -133,50 +132,56 @@ informative:
...
+  - Removed ML-KEM ciphertext from KDF.
+  - Removed RSA-KEM ciphertext from KDF.
+  - Removed references to I-D.ounsworth-lamps-cms-dhkem since we'll just inline it and don't need a DHKEM wrapper.
+
 
 Still to do in a future version:
 
   `[ ]` We need PEM samples … 118 hackathon? OQS friends? David @ BC? The right format for samples is probably to follow the hackathon ... a Dilithium or ECDSA trust anchor certificate, a composite KEM end entity certificate, and a CMS EnvolepedData sample encrypted for that composite KEM certificate.
-
+  `[ ]` Open question: do we need to incluede the ECDH, X25519, X448, and RSA public keys is the KDF? X-Wing does, but previous versions of this spec do not.
```
dconnolly: This is a good question, not sure yet

draft-ietf-lamps-pq-composite-kem.md
ounsworth: Good catch. Fixed.

ounsworth:
Closes #13

draft-ietf-lamps-pq-composite-kem.md
csosto-pk: > Similarly, RSA-KEM {{I-D.ietf-lamps-rfc5990bis}} is a bijective map between the shared secret `Z` and the ciphertext `C` and so is naturally immune to ciphertext collision. This leaves only ECDH, X25519 and X448 which are not ciphertext collision resistant and for which the ciphertexts need to be included in the derivation of the shared secret.

I am not sure I get that. Why is RSA-KEM in rfc5990bis CCR?

rfc5990bis defines

```
CEK <-  random 256-bits
z <-$ [0, n);
c <- z^e mod n;
WK <- ENCRYPT(KDF(z), CEK);
CC <- (WK, c)
RSA-ENCAPS(pubkey=(e,n)) :=  (CC, CEK) 
```

Someone that can break `c` or `WK` can produce a different `CC'` which leads to the same `CEK` by producing a different value of  `c`. ML-KEM did not have that issue because there is only one random `m` used for the encapsulation.

dstebila: I'm just looking at this now, so apologies if I'm mistaken. But I think the argument that the mapping between RSA ciphertext and plaintext is bijective only holds once you fix the modulus.  So I think you would need some binding to the modulus in the KDF input to achieve bijectivity.  (But do not take this as a proof that it's okay if you have this -- would need more careful consideration than I've given it so far.)

ounsworth:
Should align with the openpgp-pqc draft:

~~~
KEK = SHA3-256(counter || ecdhKeyShare || mlkemKeyShare || ecdhCipherText ||
 ecdhPublicKey  || mlkemCipherText || mlkemPublicKey || domSep || len(domSep) 
 || algID )
~~~

See parallel github issue: 
https://github.com/openpgp-pqc/draft-openpgp-pqc/issues/127

draft-ietf-lamps-pq-composite-kem.md
ounsworth: To err on the safe side, this has been changed to:

> Ciphertext collision resistance is not guaranteed for either RSA-KEM or ECDH, therefore these ciphertexts are bound to the key derivation.

ounsworth:
TODO: this is a massive PR. Review again when I'm less sleep-deprived.

ounsworth:
Closes #27 
Closes #2 
Closes #26

ounsworth:
Closes #2

ounsworth:
Closes #26

draft-ietf-lamps-pq-composite-kem.md
ounsworth: Isn't the definition of CCR that we care about here that you can't have `z -> c` and `z' -> c` where `z != z'` ? Since we're not borrowing the symmetric WRAP primitive from 5990, I don't see why that matters?

PR #15, opened 2024-02-06
Added a Use In CMS section

### LINKED ISSUE #12: Add a section "Use in CMS"
This draft should define how you take a composite KEM and stuff it into a KEMRI.

ounsworth:
Adding this section here avoids the need for a wrapper draft telling you how to use composite KEM in CMS.

Closes #12

### DIFF
--- a/Composite-KEM-2023.asn
+++ b/Composite-KEM-2023.asn
@@ -94,6 +94,7 @@ kema-CompositeKEM {
          VALUE CompositeCiphertextValue
          PARAMS ARE absent
          PUBLIC-KEYS { publicKeyType } 
+         SMIME-CAPS { IDENTIFIED BY id }
         }
 
 
@@ -281,5 +282,24 @@ kema-MLKEM1024-X448-KMAC256 KEM-ALGORITHM ::=
     kema-CompositeKEM{
       id-MLKEM1024-X448-KMAC256, 
       pk-MLKEM1024-X448-KMAC256 }
+
+
+--
+-- Expand the S/MIME capabilities set used by CMS [RFC5911]
+--
+
+SMimeCaps SMIME-CAPS ::=
+    { kema-MLKEM512-ECDH-P256-KMAC128 |
+      kema-MLKEM512-ECDH-brainpoolP256r1-KMAC128 |
+      kema-MLKEM512-X25519-KMAC128 |
+      kema-MLKEM512-RSA2048-KMAC128 |
+      kema-MLKEM512-RSA3072-KMAC128 |
+      kema-MLKEM768-ECDH-P256-KMAC256 |
+      kema-MLKEM768-ECDH-brainpoolP256r1-KMAC256 |
+      kema-MLKEM768-X25519-KMAC256 |
+      kema-MLKEM1024-ECDH-P384-KMAC256 |
+      kema-MLKEM1024-ECDH-brainpoolP384r1-KMAC256 |
+      kema-MLKEM1024-X448-KMAC256,
+      ... }
 
 END
--- a/Composite-KEM-2023.asn
+++ b/Composite-KEM-2023.asn
@@ -289,17 +289,17 @@ kema-MLKEM1024-X448-KMAC256 KEM-ALGORITHM ::=
 --
 
 SMimeCaps SMIME-CAPS ::=
-    { kema-MLKEM512-ECDH-P256-KMAC128 |
-      kema-MLKEM512-ECDH-brainpoolP256r1-KMAC128 |
-      kema-MLKEM512-X25519-KMAC128 |
-      kema-MLKEM512-RSA2048-KMAC128 |
-      kema-MLKEM512-RSA3072-KMAC128 |
-      kema-MLKEM768-ECDH-P256-KMAC256 |
-      kema-MLKEM768-ECDH-brainpoolP256r1-KMAC256 |
-      kema-MLKEM768-X25519-KMAC256 |
-      kema-MLKEM1024-ECDH-P384-KMAC256 |
-      kema-MLKEM1024-ECDH-brainpoolP384r1-KMAC256 |
-      kema-MLKEM1024-X448-KMAC256,
+    { kema-MLKEM512-ECDH-P256-KMAC128.&smimeCaps |
+      kema-MLKEM512-ECDH-brainpoolP256r1-KMAC128.&smimeCaps |
+      kema-MLKEM512-X25519-KMAC128.&smimeCaps |
+      kema-MLKEM512-RSA2048-KMAC128.&smimeCaps |
+      kema-MLKEM512-RSA3072-KMAC128.&smimeCaps |
+      kema-MLKEM768-ECDH-P256-KMAC256.&smimeCaps |
+      kema-MLKEM768-ECDH-brainpoolP256r1-KMAC256.&smimeCaps |
+      kema-MLKEM768-X25519-KMAC256.&smimeCaps |
+      kema-MLKEM1024-ECDH-P384-KMAC256.&smimeCaps |
+      kema-MLKEM1024-ECDH-brainpoolP384r1-KMAC256.&smimeCaps |
+      kema-MLKEM1024-X448-KMAC256.&smimeCaps,
       ... }
 
 END
--- a/draft-ietf-lamps-pq-composite-kem.md
+++ b/draft-ietf-lamps-pq-composite-kem.md
@@ -95,12 +95,12 @@ normative:
       org: "American National Standards Institute"
     date: 2007
     seriesinfo: American National Standard X9.44
-  # SHA3:
-  #   title: "SHA-3 Standard: Permutation-Based Hash and Extendable-Output Functions, FIPS PUB 202, DOI 10.6028/NIST.FIPS.202"
-  #   author:
-  #     org: "National Institute of Standards and Technology (NIST)"
-  #   date: August 2015
-  #   target: https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.202.pdf
+  SHA3:
+    title: "SHA-3 Standard: Permutation-Based Hash and Extendable-Output Functions, FIPS PUB 202, DOI 10.6028/NIST.FIPS.202"
+    author:
+      org: "National Institute of Standards and Technology (NIST)"
+    date: August 2015
+    target: https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.202.pdf
   SP800-185:
     title: "SHA-3 Derived Functions: cSHAKE, KMAC, TupleHash and ParallelHash"
     author:
@@ -143,6 +143,7 @@ informative:
   RFC2986:
   RFC4210:
   RFC4211:
+  RFC5083:
   RFC5639:
   RFC5914:
   RFC6090:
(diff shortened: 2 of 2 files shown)


draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -589,7 +567,62 @@ As with the other composite KEM algorithms, when `id-MLKEM512-RSA2048-KMAC128` o
 where:
 
 * `kda-kdf3` is defined in {{I-D.ietf-lamps-rfc5990bis}} which references it from [ANS-X9.44].
+* `id-sha3-256` is defined in {{I-D.housley-lamps-cms-sha3-hash}} which references it from [SHA3].
+
+
+# Use in CMS
+
+Composite KEM algorithms MAY be employed for one or more recipients in the CMS enveloped-data content type [RFC5652], the CMS authenticated-data content type [RFC5652], or the CMS authenticated-enveloped-data content type [RFC5083]. In each case, the KEMRecipientInfo [I-D.ietf-lamps-cms-kemri] is used with with the chosen composite KEM Algorithm to securely transfer the content-encryption key from the originator to the recipient.
```
job: Typo: “with with”

ounsworth: Thanks. Fixed.

ounsworth:
Note-to-self: look at whether we need to specify higher levels of KDF and WRAP for the stronger algorithms -- see how DVG did it in the CMS-Kyber draft.

UPDATE: done.

ounsworth:
Question: does the `## Certificate Conventions` section collide with the recently-merged Key Usage section?

UPDATE: I think this is fine because this section is intended to be able to be standalone if LAMPS wants the alg definitions, and the "Use in CMS" to become separate documents.

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -589,7 +568,92 @@ As with the other composite KEM algorithms, when `id-MLKEM512-RSA2048-KMAC128` o
...
+
+`version` is the syntax version number; it MUST be 0.
+
+`rid` identifies the recipient's certificate or public key.
+
+`kem` identifies the KEM algorithm; it MUST contain one of the OIDs listed in {{tab-kem-algs}}.
+
+`kemct` is the ciphertext produced for this recipient; it contains the `ct` output from `Encaps(pk)`.
+
+`kdf` identifies the key-derivation function (KDF). Note that the KDF used for CMS RecipientInfo process MAY be different than the KDF used within the composite KEM Algorithm.
```
```suggestion
`kdf` identifies the key-derivation function (KDF). Note that the KDF used for CMS RecipientInfo process MAY be different than the KDF used within the composite KEM Algorithm, which MAY be different than the KDFs (if any) used within the component KEMs of the composite KEM Algorithm.
```

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -589,7 +568,92 @@ As with the other composite KEM algorithms, when `id-MLKEM512-RSA2048-KMAC128` o
...
+
+The conventions specified in this section augment RFC 5280 [RFC5280].
+
+The willingness to accept a composite KEM Algorithm MAY be signaled by the use of the SMIMECapabilities Attribute as specified in Section 2.5.2. of [RFC8551] or the SMIMECapabilities certificate extension as specified in [RFC4262].
+
+The intended application for the public key MAY be indicated in the key usage certificate extension as specified in Section 4.2.1.3 of [RFC5280]. If the keyUsage extension is present in a certificate that conveys a composite KEM public key, then the key usage extension MUST contain only the following value:
+
+keyEncipherment
+
+The digitalSignatrure and dataEncipherment values MUST NOT be present. That is, a public key intended to be employed only with a composite KEM algorithm MUST NOT also be employed for data encryption or for digital signatures. This requirement does not carry any particular security consideration; only the convention that KEM keys be identifed with the `keyEncipherment` key usage.
```
```suggestion
The digitalSignature and dataEncipherment values MUST NOT be present. That is, a public key intended to be employed only with a composite KEM algorithm MUST NOT also be employed for data encryption or for digital signatures. This requirement does not carry any particular security consideration; only the convention that KEM keys be identifed with the `keyEncipherment` key usage.
```

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -589,7 +568,92 @@ As with the other composite KEM algorithms, when `id-MLKEM512-RSA2048-KMAC128` o
...
+| id-MLKEM512-X25519-KMAC128                | id-alg-hkdf-with-sha3-256 | id-aes128-Wrap   |
+| id-MLKEM512-RSA2048-KMAC128               | id-alg-hkdf-with-sha3-256 | id-aes128-Wrap   |
+| id-MLKEM512-RSA3072-KMAC128               | id-alg-hkdf-with-sha3-256 | id-aes128-Wrap   |
+| id-MLKEM768-ECDH-P256-KMAC256             | id-alg-hkdf-with-sha3-384 | id-aes192-Wrap   |
+| id-MLKEM768-ECDH-brainpoolP256r1-KMAC256  | id-alg-hkdf-with-sha3-384 | id-aes192-Wrap   |
+| id-MLKEM768-X25519-KMAC256                | id-alg-hkdf-with-sha3-384 | id-aes192-Wrap   |
+| id-MLKEM1024-ECDH-P384-KMAC256            | id-alg-hkdf-with-sha3-512 | id-aes256-Wrap   |
+| id-MLKEM1024-ECDH-brainpoolP384r1-KMAC256 | id-alg-hkdf-with-sha3-512 | id-aes256-Wrap   |
+| id-MLKEM1024-X448-KMAC256                 | id-alg-hkdf-with-sha3-512 | id-aes256-Wrap   |
+{: #tab-cms-kdf-wrap title="RECOMMENDED pairings for CMS KDF and WRAP"}
```
```suggestion
{: #tab-cms-kdf-wrap title="REQUIRED pairings for CMS KDF and WRAP"}
```

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -589,7 +568,92 @@ As with the other composite KEM algorithms, when `id-MLKEM512-RSA2048-KMAC128` o
...
+
+where:
```
danvangeest, line 656:
```suggestion

\[EDNOTE: OIDs for KMAC-based KDFs are expected. Should they be used in place of the HKDF-with-sha3 OIDs above?]

where:
```
I assume this because SHA3 isn't great to use with HKDF, and implementers will need KMAC KDF implementations anyways.

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -589,7 +568,92 @@ As with the other composite KEM algorithms, when `id-MLKEM512-RSA2048-KMAC128` o
...
+
+* `id-alg-hkdf-with-sha3-*` are defined in {{I-D.ietf-lamps-cms-sha3-hash}}.
+* `id-aes*-Wrap` are defined in [RFC3394].
+
+Implementors MAY safely substutite stronger KDF and WRAP algorithms than those indicated; for example `id-alg-hkdf-with-sha3-512` and `id-aes256-Wrap` MAY be safely used in place of `id-alg-hkdf-with-sha3-384`and `id-aes192-Wrap`, for example, where SHA3-384 or AES-192 are not supported.
```
```suggestion
Implementors MAY safely substitute stronger KDF and WRAP algorithms than those indicated; for example `id-alg-hkdf-with-sha3-512` and `id-aes256-Wrap` MAY be safely used in place of `id-alg-hkdf-with-sha3-384`and `id-aes192-Wrap`, for example, where SHA3-384 or AES-192 are not supported.
```

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -589,7 +568,92 @@ As with the other composite KEM algorithms, when `id-MLKEM512-RSA2048-KMAC128` o
...
+
+EDNOTE: The convention in LAMPS is to specify algorithms and their CMS conventions in separate documents. Here we have presented them in the same document, but this section has been written so that it can easily be moved to a standalone document.
+
+Composite KEM algorithms MAY be employed for one or more recipients in the CMS enveloped-data content type [RFC5652], the CMS authenticated-data content type [RFC5652], or the CMS authenticated-enveloped-data content type [RFC5083]. In each case, the KEMRecipientInfo [I-D.ietf-lamps-cms-kemri] is used with the chosen composite KEM Algorithm to securely transfer the content-encryption key from the originator to the recipient.
+
+## Underlying Components
+
+A CMS implementation that supports a composite KEM algorithm MUST support at least the following underlying components:
+
+For the key-derivation function, an implementation MUST support `id-alg-hkdf-with-sha3-256` {{I-D.ietf-lamps-cms-sha3-hash}}.
```
```suggestion
When a particular Composite KEM OID is supported, an implementation MUST support the corresponding KDF algorithm identifier in {{tab-cms-kdf-wrap}}.
```
id-alg-hkdf-with-sha3-256 is too low a bar for NIST level 5 KEMs.

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -589,7 +568,92 @@ As with the other composite KEM algorithms, when `id-MLKEM512-RSA2048-KMAC128` o
...
+
+For key-wrapping, an implementation MUST support the AES-Wrap-128 [RFC3394] key-encryption algorithm.
```
```suggestion
When a particular Composite KEM OID is supported, an implementation MUST support the corresponding WRAP algorithm identifier in {{tab-cms-kdf-wrap}}.
```
AES-Wrap-128 is too low a bar for NIST level 5 KEMs.

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -589,7 +568,92 @@ As with the other composite KEM algorithms, when `id-MLKEM512-RSA2048-KMAC128` o
...
+
+An implementation MAY also support other key-derivation functions and other key-encryption algorithms as well.
+
+The following table lists the RECOMMENDED KDF and WRAP algorithms to preserve security and performance characteristics of each composite algorithm.
```
```suggestion
The following table lists the REQUIRED KDF and WRAP algorithms to preserve security and performance characteristics of each composite algorithm.
```

ounsworth:
We also need to add the SMIME-CAPS to the ASN.1 (find a good example doc).

danvangeest (commented):
Sorry! I created this review last week but forgot to actually submit it!

ounsworth:
Thanks for the review @danvangeest 

Still TODO:

* We also need to add the SMIME-CAPS to the ASN.1 (find a good example doc).
  * UPDATE: done.

ounsworth:
Review comments from Russ:

> Mike:
>
>You are using a mix of "wrap algorithm" and "key-encryption algorithm".  CMS usually uses the later.  Line 683 got it just right.  That is, id-aes256-Wrap identifies a particular key-encryption algorithm.  I am not sure the text at the top is aligned.
>
>Line 663: I think you need o say how that happens.  In some cases a different OID is used, but in others a different parameter value is used.
>
>Line 675: make it clear thet it is the KEM.Encaps(pk).  The function is specific to the KEM identified in 'kem'.
>
>Line 681: Do any of your composite KEM algorithms REQUIRE a ukm?  I can imagine a way to use DH and ECDH that require a UKM.
>
>Russ

Composite-KEM-2023.asn
```diff
@@ -282,4 +283,23 @@ kema-MLKEM1024-X448-KMAC256 KEM-ALGORITHM ::=
       id-MLKEM1024-X448-KMAC256, 
       pk-MLKEM1024-X448-KMAC256 }
 
+
+--
+-- Expand the S/MIME capabilities set used by CMS [RFC5911]
+--
+
+SMimeCaps SMIME-CAPS ::=
+    { kema-MLKEM512-ECDH-P256-KMAC128 |
```
danvangeest: I don't know anything about ASN.1 but shouldn't these be `kema-MLKEM512-ECDH-P256-KMAC128.&smimeCaps`, etc.  I just base this on what I see in RFC 5911.

ounsworth: Yup, sure should. Good catch.

PR #30, opened 2024-07-07
Lifted AES-192 to AES-256. Closes #25

### LINKED ISSUE #25: Consider lifting AES-192 to AES-256
In cms-kyber, we specified id-aes256-wrap for NIST level 3 / 192 bit security, even though it's stronger than necessary, because of a suggestion from Scott that AES-192 isn't as widely supported.  Consider doing the same here.

ounsworth:


johngray-dev:
Thanks Mike.  Edits look really great and justification text is also good.

PR #32, opened 2024-07-07
Replaced RSA-KEM with RSA-OAEP

### LINKED ISSUE #21: Should we instead or in addition support RSA-OAEP?
Sophie Schmieg made the excellent point:

> on the classical side, nobody uses RSA-KEM (a shame, but alas), leaving RSA-OAEP

### LINKED ISSUE #24: Is section "## RSA-KEM Parameters" actually needed?
Go read 5990bis more carefully.

Also, maybe not needed at all if we go with RSA-OAEP.

ounsworth:
Closes #21 
Closes #24

### DIFF
--- a/draft-ietf-lamps-pq-composite-kem.md
+++ b/draft-ietf-lamps-pq-composite-kem.md
@@ -76,29 +76,30 @@ author:
 normative:
   RFC2119:
   RFC3394:
+  RFC3560:
+  RFC4055:
   RFC5280:
   RFC5652:
   RFC5958:
   RFC8174:
   RFC8410:
   RFC8411:
-  I-D.draft-ietf-lamps-rfc5990bis-04:
   I-D.draft-ietf-lamps-cms-kemri-08:
   I-D.draft-ietf-lamps-cms-sha3-hash-04:
-  ANS-X9.44:
-    title: "Public Key
-              Cryptography for the Financial Services Industry -- Key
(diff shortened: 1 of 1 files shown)


draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -320,16 +331,35 @@ We borrow here the definition of a key encapsulation mechanism (KEM) from {{I-D.
...
 
 The `KeyGen() -> (pk, sk)` of a composite KEM algorithm will perform the `KeyGen()` of the respective component KEM algorithms and it produces a composite public key `pk` as per {{sec-composite-pub-keys}} and a composite secret key `sk` is per {{sec-priv-key}}.
 
-### Promotion of RSA into a KEM
+### Promotion of RSA-OAEP into a KEM
+
+The RSA Optimal Asymmetric Encription Padding (OAEP), more specifically the RSAES-OAEP key transport algorithm as specified in [RFC3560] is a public key encryption algorithm used to transport key material from a sender to a reviever. It is promoted into a KEM by having the sender generat a randem 256 bit secret and encrypt it.
+
+~~~
+DHKEM.Encaps(pkR):
```
dstebila, line 346: Why is this title `DHKEM.Encaps` -- shouldn't it be something like `RSAOAEPKEM.Encaps`?

PR #56, opened 2024-08-06
Editorial changes

ounsworth:


draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -930,20 +934,9 @@ The term "ease of migration" is used here to mean that existing systems can be g
...
-
-EDNOTE: I copied and pruned this text from I-D.ounsworth-pq-composite-sigs. It probably needs to be fleshed out more as we better understand the implementation concerns around composite encryption.
+ML-KEM always requires the public key in order to perform various steps of the Fujisaki-Okamoto decapsulation [FIPS203-ipd]. Additionally, in order to achaive the public-key binding property the KEM combiner used to form the composite KEM requires the public key as input to the KDF that derives the output shared secret. One way to carry the public key would be to include each component's public key within the respective OneAsymmetricKey as is required by the private key encoding given in {{sec-priv-key}}. Implementers who choose a different private key encoding MUST consider how to private the component public keys to the decapsulate routine. While some implementations might contain routines to computationally derive the public key from the private key, it is not guaranteed that all implementations will support this; for this reason the interoparable composite private key format given in this document in {{sec-priv-key}} requires the component public keys to be included.
```
ounsworth: typo "achaive".

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -500,7 +504,7 @@ The parameters field MUST be absent.
...
 
-When a `CompositePrivateKey` is conveyed inside a OneAsymmetricKey structure (version 1 of which is also known as PrivateKeyInfo) [RFC5958], the privateKeyAlgorithm field SHALL be set to the corresponding composite algorithm identifier defined according to {{sec-alg-ids}}, the privateKey field SHALL contain the CompositeKEMPrivateKey, and the publicKey field MUST NOT be present. Associated public key material MAY be present in the CompositeKEMPrivateKey.
+When a `CompositePrivateKey` is conveyed inside a OneAsymmetricKey structure (version 1 of which is also known as PrivateKeyInfo) [RFC5958], then a situation arises where multiple `OneAsymmetricKey`s are nested inside an outer `OneAsymmetricKey`, which  the privateKeyAlgorithm field SHALL be set to the corresponding composite algorithm identifier defined according to {{sec-alg-ids}}, the privateKey field SHALL contain the CompositeKEMPrivateKey, and the publicKey field MUST NOT be present. Associated public key material MUST be present in the CompositeKEMPrivateKey. The reason for requiring component public keys to be carried with the private keys is discussed in {{impl-cons-decaps-pubkey}}
```
ounsworth: We should think about this more. ML-KEM's FO requires the public key anyway, so the ML-KEM's private key needs to have it anyway, so we don't need to duplicate it at the composite layer. Same for RSA: the public key is a subset of the private key. We only really need to think about ECC (RFC 5915). Perhaps we need to treat ECC as a special case here?

Certainly, the "MUST" needs to be broken out into cases.

ounsworth:
The security considerations need more work, especially the one about key reuse.

Another argument against key reuse is documented in https://github.com/lamps-wg/draft-composite-sigs/issues/49

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -156,7 +156,8 @@ informative:
   RFC8551:
   RFC9180:
   I-D.draft-ietf-tls-hybrid-design-04:
-  I-D.draft-driscoll-pqt-hybrid-terminology-01:
+  I-D.draft-ietf-pquip-pqt-hybrid-terminology-03:
```
janklaussner: Its already -04

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -704,35 +730,31 @@ Composite KEM algorithms MAY be employed for one or more recipients in the CMS e
...
+| id-MLKEM512-RSA3072               | id-sha3-256 | id-aes128-Wrap     |
+| id-MLKEM768-ECDH-P256             | id-sha3-384 | id-aes256-Wrap     |
+| id-MLKEM768-ECDH-brainpoolP256r1  | id-sha3-384 | id-aes256-Wrap     |
+| id-MLKEM768-X25519                | id-sha3-384 | id-aes256-Wrap     |
+| id-MLKEM1024-ECDH-P384            | id-sha3-512 | id-aes256-Wrap     |
+| id-MLKEM1024-ECDH-brainpoolP384r1 | id-sha3-512 | id-aes256-Wrap     |
+| id-MLKEM1024-X448                 | id-sha3-512 | id-aes256-Wrap     |
 {: #tab-cms-kdf-wrap title="REQUIRED pairings for CMS KDF and WRAP"}
 
+In all cases above, the hash function used as a KDF produces a longer output than needed by the encryption algorithm, so the output should simply be truncated to the correct length -- ie use the leftmost significant bits.
```
janklaussner: Is there a reason why it says „should“ and not „shall“ or „must“? Proposal: 
[..] output shall be truncated […] by using the leftmost bits.

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -930,20 +976,12 @@ The term "ease of migration" is used here to mean that existing systems can be g
...
 
-EDNOTE: I copied and pruned this text from I-D.ounsworth-pq-composite-sigs. It probably needs to be fleshed out more as we better understand the implementation concerns around composite encryption.
+With regard to the traditional algorithms, RSA or Elliptic CUrve, in order to achieve the public-key binding property the KEM combiner used to form the composite KEM, the combiner requires the traditional public key as input to the KDF that derives the output shared secret. Therefore it is required to carry the public key within the respective `OneAsymmetricKey.publicKey` as per the private key encoding given in {{sec-priv-key}}. Implementers who choose to use a different private key encoding than the one specified in this document MUST consider how to provide the component public keys to the decapsulate routine. While some implementations might contain routines to computationally derive the public key from the private key, it is not guaranteed that all implementations will support this; for this reason the interoparable composite private key format given in this document in {{sec-priv-key}} requires the public key of the traditional component to be included.
```
janklaussner: Typo: „CUrve“

janklaussner (changes requested):
2 small issues, one question

draft-ietf-lamps-pq-composite-kem.md
ounsworth: Done

ounsworth:
Closes #72

draft-ietf-lamps-pq-composite-kem.md
ounsworth: Done

ounsworth:
This PR has not been fully reviewed by the author's group, but I have been over it several times and I am merging it to close off all the related Issues.

PR #57, opened 2024-08-10
RSA-OAEP tweaks.

### LINKED ISSUE #35: Russ feedback on RSA-OAEP
https://mailarchive.ietf.org/arch/msg/spasm/e7OlkcpDMtOWbGLPQB1hEPyyjXQ/

---
Mike:

I think you should reference Section 7.1 of RFC 8017 for the RSA-OAEP specification.

Then, the description should use RSAES-OAEP-ENCRYPT and RSAES-OAEP-DECRYPT as specified there.

I assume that you are thinking about the L parameter being an empty string.

The Mask Generation Function needs to be specified.  I assume you want to use MGF1 from Appendix B.2.1 of RFC 8017.  I think you can state the maskLen for the two cases that you are supporting.

Russ

ounsworth:
Closes #35

### DIFF
--- a/draft-ietf-lamps-pq-composite-kem.md
+++ b/draft-ietf-lamps-pq-composite-kem.md
@@ -221,8 +221,7 @@ This document introduces a set of Key Encapsulation Mechanism (KEM) schemes that
 # Changes in version -05
 
 * Fixed a bug in the definition of the Encaps() functions: KEMs, according to both RFC9180 and FIPS 203 should always return (ss, ct), but we had (ct, ss).
-* Adjusted RSA-OAEP section to follow RFC8017 instead of RFC3560.
-  * OPEN QUESTION: RSAES-OAEP in RFC8017 can take a label. Should we use it? Perhaps to pass in the same domain separator as used for the overall composite?
+* Adjusted RSA-OAEP section to follow RFC8017 instead of RFC3560. Does not use the RSA-OAEP label.
 
 
 Still to do in a future version:
--- a/draft-ietf-lamps-pq-composite-kem.md
+++ b/draft-ietf-lamps-pq-composite-kem.md
@@ -369,9 +369,7 @@ DHKEM.Encaps(pkR):
   return shared_secret, enc
 ~~~
 
-EDNOTE:
-
- `Decaps(sk, ct) -> ss` is accomplished in the analogous way.
+`Decaps(sk, ct) -> ss` is accomplished in the analogous way.
 
 ~~~
 DHKEM.Decap(skR, enc):
@@ -659,8 +657,6 @@ EDNOTE: I believe that [SP.800-56Ar3] and [BSI-ECC] give equivalent and inter-op
 The KEM combiner defined in section {{sec-kem-combiner}} requires a domain separator `domSep` input.  The following table shows the HEX-encoded domain separator for each Composite KEM AlgorithmID; to use it, the value should be HEX-decoded and used in binary form. The domain separator is simply the DER encoding of the composite algorithm OID.
 
 EDNOTE: Should the domain separator values be the SHA-256 hash of the DER encoding of the corresponding composite algorithm OID? That way they would be fixed-length even if the OIDs are different lengths. See https://github.com/lamps-wg/draft-composite-sigs/issues/19
-
-EDNOTE: Or should the domain separators be synced up with the corresponding OpenPGP draft?
 
 | Composite KEM AlgorithmID | Domain Separator (in Hex encoding)|
 | ----------- | ----------- |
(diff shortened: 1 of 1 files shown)


draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -330,26 +331,29 @@ CompositeKEM.KeyGen():
...
 ### Promotion of RSA-OAEP into a KEM
 
-The RSA Optimal Asymmetric Encryption Padding (OAEP), more specifically the RSAES-OAEP key transport algorithm as specified in [RFC3560] is a public key encryption algorithm used to transport key material from a sender to a receiver. It is promoted into a KEM by having the sender generate a random 256 bit secret and encrypt it.
+The RSA Optimal Asymmetric Encryption Padding (OAEP), as defined in section 7.1 of [RFC8017] is a public key encryption algorithm used to transport key material from a sender to a receiver. It is promoted into a KEM by having the sender generate a random 256 bit secret and encrypt it.
 
 ~~~
 RSAOAEPKEM.Encaps(pkR):
   shared_secret = SecureRandom(ss_len)
-  enc = RSA-OAEP.Encrypt(pkR, shared_secret)
+  enc = RSAES-OAEP-ENCRYPT(pkR, shared_secret, domSep)
```
ounsworth: We do not believe that passing domSep down into RSA-OAEP does not give any security advantage, so it would simplify implementations if we leave OAEP's `L` empty because empty is the standard way to implement PKCS#1 OAEP.

Change to empty string.

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -673,27 +681,30 @@ EDNOTE: these domain separators are based on the prototyping OIDs assigned on th
...
-| hashFunc                  | id-sha2-256                   |
-| maskGenFunc               | mgf1SHA256Identifier          |
-| pSourceFunc           | DEFAULT pSpecifiedEmptyIdentifier |
-| ss_len                    | 256 bits                      |
+| RSAES-OAEP-params           | Value                       |
+| ----------------------      | ---------------             |
+| hashAlgorithm               | id-sha2-256                 |
+| maskGenAlgorithm            | mgf1SHA256Identifier        |
+| pSourceAlgorithm            | id-pSpecified               |
+| PSourceALgorithm.parameters | domSep                      |
```
ounsworth: Change to empty string, ie fixed value as per 8017.

ounsworth: Also change id-pSpecified to id-empty or whatever.

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -330,26 +330,29 @@ CompositeKEM.KeyGen():
 
 ### Promotion of RSA-OAEP into a KEM
 
-The RSA Optimal Asymmetric Encryption Padding (OAEP), more specifically the RSAES-OAEP key transport algorithm as specified in [RFC3560] is a public key encryption algorithm used to transport key material from a sender to a receiver. It is promoted into a KEM by having the sender generate a random 256 bit secret and encrypt it.
+The RSA Optimal Asymmetric Encryption Padding (OAEP), as defined in section 7.1 of [RFC8017] is a public key encryption algorithm used to transport key material from a sender to a receiver. It is promoted into a KEM by having the sender generate a random 256 bit secret and encrypt it.
 
 ~~~
 RSAOAEPKEM.Encaps(pkR):
```
ounsworth: To match 9180, this should be "Encap()" and "Decap()".
TODO: change throughout the document.

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -673,27 +676,29 @@ EDNOTE: these domain separators are based on the prototyping OIDs assigned on th
...
 
 where:
 
 * `id-sha256` is defined in [RFC8017].
-* `mgf1SHA256Identifier` is defined in [RFC4055].
-* `pSpecifiedEmptyIdentifier` is defined in [RFC3560]
+* `mgf1SHA256Identifier` is defined in [RFC4055], which refers to the MFG1 function defined in [RFC8017] appendix B.2.1.
+* `pSpecifiedEmpty` is defined in [RFC8017] to indicate that the empty string is used for the label.
+
+EDNOTE: we could determine the mask length for each parameter set given in this document. According to 8017, it will be the length of `k - hLen - 1`, where `k` is the size of the RSA modulus. I'm not sure that we strictly need to calculate and list these.
```
ounsworth: Turn this into a real note.
"... since the mask length will be constant for each composite algorithm, implementations could pre-compute and hard-code the mask length value."

ounsworth:
After I make the two changes noted above, I am ok to go ahead and merge this.

PR #76, opened 2024-10-14
Add explicit ALgorithmIdentifiers of component keys.

### LINKED ISSUE #20: Add a new section: explicitly list SPKI AlgIds
We should add a section listing explicitly the DER-encoded AlgorithmIdentifiers for the components of each composite public key and signature algorithm. This is important to resolve ambiguity on, for example, whether the RSA should have a NULL param, and the ECC curve params.

Example, for id-MLDSA44-ECDSA-P256-SHA256 the ML-DSA SPKI would have an AlgorithmIdentifier of:

   AlgorithmIdentifier  ::=  SEQUENCE  {
        id-ml-dsa
    }
which is:

AlgorithmIdentifier  ::=  SEQUENCE  {
        {
     1.3.6.1.4.1.2.267.12.4.4
      }
  }
And the ECDSA-P256-SHA256 would have a SPKI would have an AlgorithmIdentifier of:

   AlgorithmIdentifier  ::=  SEQUENCE  {
        id-ecPublicKey,
       secp256r1  
  }
which is:

   AlgorithmIdentifier  ::=  SEQUENCE  {
        {
     iso(1) member-body(2) us(840) ansi-X9-62(10045) keyType(2) 1 },
       iso(1) member-body(2) us(840) ansi-X9-62(10045) curves(3) prime(1) 7}
And the signature algorithm for id-MLDSA44-ECDSA-P256-SHA256, the first component signature algorithm would have an AlgorithmIdentifier of

   AlgorithmIdentifier  ::=  SEQUENCE  {
        id-ml-dsa
    }
which is:

AlgorithmIdentifier  ::=  SEQUENCE  {
        {
     1.3.6.1.4.1.2.267.12.4.4
      }
  }
and the second component signature algorithm would have an AlgorithmIdentifier of

   AlgorithmIdentifier  ::=  SEQUENCE  {
          ecdsa-with-SHA256
  }
which is:

AlgorithmIdentifier  ::=  SEQUENCE  {
           {
     iso(1) member-body(2) us(840) ansi-X9-62(10045) signatures(4) ecdsa-with-SHA2(3) 2
       }
  }
With that done, we should replace the message prefix values in Sectien 2.4 with the SHA256 hash of the signature AlgorithmIdentifiers. This has two nice properties that are better than using the ASCII encoding of the OID name: 1) they are all the same length (ie the length of SHA256), and 2) if the inner OIDs change, for example with a new Kyber version, then the message prefix changes, which prevents cryptographic compatibility issues; or otherwise stated: provides signature domain-separation based on the component OIDs.

--- SHA256 of the DER encoding of the following ASN.1 value
--- Security Consideration note: the choice of SHA256 here is not security-relevant since it is only to generate fixed string values.

SEQUENCE {
AlgorithmIdentifier  ::=  SEQUENCE  {
        {
     1.3.6.1.4.1.2.267.12.4.4
      }
  },
AlgorithmIdentifier  ::=  SEQUENCE  {
           {
     iso(1) member-body(2) us(840) ansi-X9-62(10045) signatures(4) ecdsa-with-SHA2(3) 2
       }
  }
}

janklaussner:


janklaussner:
Question to group: for RSA key AlgIds it is possible to be very specific on its usage. Do we want/need this?

ounsworth:
This does not yet provide an encoding of OAEP parameters. We need to address that before merging.

Update: Just need John to provide a DER encoding of the RSA-OAEP params.

janklaussner:
@ounsworth , @johngray-dev : added DER encoding and some places where RSA 4096 combo was missing... I assumed that the OAEP params for RSA 4096 are the same as for RSA 2048

ounsworth:
Thanks Jan! I'm going to merge this now.

ounsworth (approved):

ounsworth:
@janklaussner your DER encoding of the RSA-OAEP AlgID had an extra NULL at the end that John and I couldn't figure out, so I replaced it with a DER encoding produced by our toolkit.

janklaussner:
> @janklaussner your DER encoding of the RSA-OAEP AlgID had an extra NULL at the end that John and I couldn't figure out, so I replaced it with a DER encoding produced by our toolkit.

@ounsworth @johngray-dev I'm using OpenSSL and try to create or import OAEP keys, but neither my nor your encoding worked. How did you do it? Can you send me an example public key that works for you? Otherwise I doubt this kind of "key" is supported by OpenSSL.
Thanks!

PR #78, opened 2024-10-18
Aligned Use In CMS section with draft-cms-kyber

### LINKED ISSUE #65: Align CMS KDF choice to Dan vG's cms-kyber draft
The current CMS KEMRI Conventions:

| Composite KEM OID                 | KDF         | Key Encryption Alg |
|---------                          | ---         | ---                |
| id-MLKEM768-RSA2048               | SHA3-256 | id-aes128-Wrap     |
| id-MLKEM768-RSA3072               | SHA3-256 | id-aes128-Wrap     |
| id-MLKEM768-ECDH-P384             | SHA3-384 | id-aes256-Wrap     |
| id-MLKEM768-ECDH-brainpoolP256r1  | SHA3-384 | id-aes256-Wrap     |
| id-MLKEM768-X25519                | SHA3-384 | id-aes256-Wrap     |
| id-MLKEM1024-ECDH-P384            | SHA3-512 | id-aes256-Wrap     |
| id-MLKEM1024-ECDH-brainpoolP384r1 | SHA3-512 | id-aes256-Wrap     |
| id-MLKEM1024-X448                 | SHA3-512 | id-aes256-Wrap     |

Which does not currently align with the KDFs used internally within the Composite KEM Combiner, and in fact here it needs to be a proper variable-length KDF since KEMRI has an `L' parameter, so this needs to be fixed. I guess it makes sense for me to align to whatever Dan picks in draft-cms-kyber.

ounsworth:
Closes #65

### DIFF
--- a/draft-ietf-lamps-pq-composite-kem.md
+++ b/draft-ietf-lamps-pq-composite-kem.md
@@ -774,7 +774,7 @@ All recommendations for using Composite ML-KEM in CMS are fully aligned with the
 
 ## RecipientInfo Conventions {#sec-using-recipientInfo}
 
-When the ML-KEM algorithm is employed for a recipient, the RecipientInfo alternative for that recipient MUST be OtherRecipientInfo using the KEMRecipientInfo structure as defined in {{RFC9629}}.
+When Composite ML-KEM is employed for a recipient, the RecipientInfo alternative for that recipient MUST be OtherRecipientInfo using the KEMRecipientInfo structure as defined in {{RFC9629}}.
 
 The fields of the KEMRecipientInfo MUST have the following values:
 
--- a/draft-ietf-lamps-pq-composite-kem.md
+++ b/draft-ietf-lamps-pq-composite-kem.md
@@ -301,7 +301,7 @@ PQ/T Hybrid cryptography can, in general, provide solutions to two migration pro
 - Algorithm strength uncertainty: During the transition period, some post-quantum signature and encryption algorithms will not be fully trusted, while also the trust in legacy public key algorithms will start to erode.  A relying party may learn some time after deployment that a public key algorithm has become untrustworthy, but in the interim, they may not know which algorithm an adversary has compromised.
 - Ease-of-migration: During the transition period, systems will require mechanisms that allow for staged migrations from fully classical to fully post-quantum-aware cryptography.
 
-This document defines a specific instantiation of the PQ/T Hybrid paradigm called "composite" where multiple cryptographic algorithms are combined to form a single key encapsulation mechanism (KEM) key and ciphertext such that they can be treated as a single atomic algorithm at the protocol level. Composite algorithms address algorithm strength uncertainty because the composite algorithm remains strong so long as one of its components remains strong. Concrete instantiations of composite KEM algorithms are provided based on ML-KEM, RSA-OAEP and ECDH. Backwards compatibility is not directly covered in this document, but is the subject of {{sec-backwards-compat}}.
+This document defines a specific instantiation of the PQ/T Hybrid paradigm called "composite" where multiple cryptographic algorithms are combined to form a single key encapsulation mechanism (KEM) key and ciphertext such that they can be treated as a single atomic algorithm at the protocol level. Composite algorithms address algorithm strength uncertainty because the composite algorithm remains strong so long as one of its components remains strong. Concrete instantiations of composite ML-KEM algorithms are provided based on ML-KEM, RSA-OAEP and ECDH. Backwards compatibility is not directly covered in this document, but is the subject of {{sec-backwards-compat}}.
 
 Composite ML-KEM is intended for general applicability anywhere that key establishment or enveloped content encryption is used within PKIX or CMS structures.
 
@@ -369,7 +369,7 @@ The KEM interface was chosen as the interface for a composite key establishment 
 
 ### Composite KeyGen
 
-The `KeyGen() -> (pk, sk)` of a composite KEM algorithm will perform the `KeyGen()` of the respective component KEM algorithms and it produces a composite public key `pk` as per {{sec-composite-pub-keys}} and a composite secret key `sk` as per {{sec-priv-key}}.
+The `KeyGen() -> (pk, sk)` of a composite ML-KEM algorithm will perform the `KeyGen()` of the respective component KEM algorithms and it produces a composite public key `pk` as per {{sec-composite-pub-keys}} and a composite secret key `sk` as per {{sec-priv-key}}.
 
 ~~~
 CompositeKEM.KeyGen():
@@ -391,7 +391,7 @@ RSAOAEPKEM.Encap(pkR):
   return shared_secret, enc
 ~~~
 
-Note that the OAEP label `L` is left to its default value, which is the empty string as per [RFC8017]. The shared secret output by the overall composite KEM already binds a composite domain separator, so there is no need to also utilize the component domain separators.
+Note that the OAEP label `L` is left to its default value, which is the empty string as per [RFC8017]. The shared secret output by the overall Composite ML-KEM already binds a composite domain separator, so there is no need to also utilize the component domain separators.
 
 The value of `ss_len` as well as the RSA-OAEP parameters used within this specification can be found in {{sect-rsaoaep-params}}.
 
@@ -433,7 +433,7 @@ DHKEM.Decap(skR, enc):
 
 This construction applies for all variants of elliptic curve Diffie-Hellman used in this specification: ECDH, X25519, and X448.
 
-The simplifications from the DHKEM definition in [RFC9180] is that since the ciphertext and receiver's public key are included explicitly in the composite KEM combiner, there is no need to construct the `kem_context` object, and since a domain separator is included explicitly in the composite KEM combiner there is no need to perform the labelled steps of `ExtractAndExpand()`.
+The simplifications from the DHKEM definition in [RFC9180] is that since the ciphertext and receiver's public key are included explicitly in the Composite ML-KEM combiner, there is no need to construct the `kem_context` object, and since a domain separator is included explicitly in the Composite ML-KEM combiner there is no need to perform the labelled steps of `ExtractAndExpand()`.
 
 
 ### KEM Combiner {#sec-kem-combiner}
@@ -456,10 +456,10 @@ where:
 * `tradSS` is the shared secret from the traditional component (elliptic curve or RSA).
 * `tradCT` is the ciphertext from the traditional component (elliptic curve or RSA).
 * `tradPK` is the public key of the traditional component (elliptic curve or RSA).
-* `domSep` SHALL be the DER encoded value of the object identifier of the composite KEM algorithm as listed in {{sec-domain}}.
+* `domSep` SHALL be the DER encoded value of the object identifier of the Composite ML-KEM algorithm as listed in {{sec-domain}}.
 * `||` represents concatenation.
 
-Each registered composite KEM algorithm specifies the choice of `KDF`, `demSep` to be used in {{sec-alg-ids}} and {{sec-domain}} below. Given that each composite KEM algorithm fully specifies the component algorithms, including for example the size of the RSA modulus, all inputs to the KEM combiner are fixed-size and thus do not require length-prefixing. The `CompositeKEM.Decap()` specified in {{sect-composite-decaps}} adds further error handling to protect the KEM combiner from malicious inputs.
+Each registered Composite ML-KEM algorithm specifies the choice of `KDF`, `demSep` to be used in {{sec-alg-ids}} and {{sec-domain}} below. Given that each Composite ML-KEM algorithm fully specifies the component algorithms, including for example the size of the RSA modulus, all inputs to the KEM combiner are fixed-size and thus do not require length-prefixing. The `CompositeKEM.Decap()` specified in {{sect-composite-decaps}} adds further error handling to protect the KEM combiner from malicious inputs.
 
 
 
@@ -467,7 +467,7 @@ Each registered composite KEM algorithm specifies the choice of `KDF`, `demSep` 
 ### Composite Encap
 Note that, at least at the time of writing, the algorithm `DHKEM` is not defined as a standalone algorithm within PKIX standards and it does not have an assigned algorithm OID, so it connot be used directly with CMS KEMRecipientInfo [RFC9629]; it is merely a building block for the composite algorithm.
 
-The `Encap(pk) -> (ss, ct)` of a composite KEM algorithm is defined as:
+The `Encap(pk) -> (ss, ct)` of a Composite ML-KEM algorithm is defined as:
 
 ~~~
 CompositeKEM.Encap(pk):
@@ -488,11 +488,11 @@ CompositeKEM.Encap(pk):
   return (ss, ct)
 ~~~
 
-where `Combiner(tradSS, mlkemSS, tradCT, tradPK, domSep)` is defined in general in {{sec-kem-combiner}} with specific values for `domSep` per composite KEM algorithm in {{sec-alg-ids}} and `CompositeCiphertextValue` is defined in {{sec-CompositeCiphertextValue}}.
+where `Combiner(tradSS, mlkemSS, tradCT, tradPK, domSep)` is defined in general in {{sec-kem-combiner}} with specific values for `domSep` per Composite ML-KEM algorithm in {{sec-alg-ids}} and `CompositeCiphertextValue` is defined in {{sec-CompositeCiphertextValue}}.
 
 ### Composite Decap {#sect-composite-decaps}
 
-The `Decap(sk, ct) -> ss` of a composite KEM algorithm is defined as:
+The `Decap(sk, ct) -> ss` of a Composite ML-KEM algorithm is defined as:
 
 ~~~
 CompositeKEM.Decap(ct, mlkemSK, tradSK):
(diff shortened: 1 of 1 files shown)


draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -782,29 +770,86 @@ Note: The mask length, according to [RFC8017], is `k - hLen - 1`, where `k` is t
 
 Composite KEM algorithms MAY be employed for one or more recipients in the CMS enveloped-data content type [RFC5652], the CMS authenticated-data content type [RFC5652], or the CMS authenticated-enveloped-data content type [RFC5083]. In each case, the KEMRecipientInfo [RFC9629] is used with the chosen composite KEM Algorithm to securely transfer the content-encryption key from the originator to the recipient.
```
danvangeest: General comment: Throughout the document, instead of saying "Composite KEM", should you say "Composite ML-KEM"?

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -782,29 +770,86 @@ Note: The mask length, according to [RFC8017], is `k - hLen - 1`, where `k` is t
...
 
+All recommendations for using CompositeML-KEM in CMS are fully aligned with the use of ML-KEM in CMS {{I-D.ietf-lamps-cms-kyber}}.
```
```suggestion
All recommendations for using Composite ML-KEM in CMS are fully aligned with the use of ML-KEM in CMS {{I-D.ietf-lamps-cms-kyber}}.
```

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -782,29 +770,86 @@ Note: The mask length, according to [RFC8017], is `k - hLen - 1`, where `k` is t
...
+
+## RecipientInfo Conventions {#sec-using-recipientInfo}
+
+When the ML-KEM algorithm is employed for a recipient, the RecipientInfo alternative for that recipient MUST be OtherRecipientInfo using the KEMRecipientInfo structure as defined in {{RFC9629}}.
```
```suggestion
When Composite ML-KEM is employed for a recipient, the RecipientInfo alternative for that recipient MUST be OtherRecipientInfo using the KEMRecipientInfo structure as defined in {{RFC9629}}.
```

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -782,29 +770,86 @@ Note: The mask length, according to [RFC8017], is `k - hLen - 1`, where `k` is t
...
+
+The fields of the KEMRecipientInfo MUST have the following values:
+
+> version is the syntax version number; it MUST be 0.
+
+> rid identifies the recipient's certificate or public key.
+
+> kem identifies the KEM algorithm; it MUST contain one of the CompositeML-KEM identifiers listed in {{sec-alg-ids}}.
```
```suggestion
> kem identifies the KEM algorithm; it MUST contain one of the Composite ML-KEM identifiers listed in {{sec-alg-ids}}.
```

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -782,29 +770,86 @@ Note: The mask length, according to [RFC8017], is `k - hLen - 1`, where `k` is t
...
+
+> kemct is the ciphertext produced for this recipient.
+
+> kdf identifies the key-derivation algorithm. Note that the Key Derivation Function (KDF) used for CMS RecipientInfo process MAY be different than the KDF used within the ML-KEM algorithm.
```
```suggestion
> kdf identifies the key-derivation algorithm. Note that the Key Derivation Function (KDF) used for CMS RecipientInfo process MAY be different than the KDF used within the Composite ML-KEM algorithm or one of its components.
```

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -782,29 +770,86 @@ Note: The mask length, according to [RFC8017], is `k - hLen - 1`, where `k` is t
...
+
+> kekLength is the size of the key-encryption key in octets.
+
+> ukm is an optional random input to the key-derivation function. ML-KEM doesn't place any requirements on the ukm contents.
```
danvangeest, line 802:
```suggestion
> ukm is an optional random input to the key-derivation function. Composite ML-KEM doesn't place any requirements on the ukm contents.
```

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -782,29 +770,86 @@ Note: The mask length, according to [RFC8017], is `k - hLen - 1`, where `k` is t
...
+info:
+: optional context and application specific information. In this document this corresponds to the info KDF input from {{Section 5 of RFC9629}}. This is the ASN.1 DER encoding of CMSORIforKEMOtherInfo.
+
+L:
+: length of output keying material in octets. This corresponds to the L KDF input from {{Section 5 of RFC9629}}, which is identified in the kekLength value from KEMRecipientInfo. Implementations MUST confirm that this value is consistent with the key size of the key-encryption algorithm.
+
+HKDF may be used with different hash functions, including SHA-256 {{FIPS.180-4}}. The object identifier id-alg-hkdf-with-sha256 is defined in {{RFC8619}}, and specifies the use of HKDF with SHA-256. The parameter field MUST be absent when this algorithm identifier is used to specify the KDF for ML-KEM in KemRecipientInfo.
+
+
+### Components for CompositeML-KEM in CMS
```
```suggestion
### Components for Composite ML-KEM in CMS
```

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -782,29 +770,86 @@ Note: The mask length, according to [RFC8017], is `k - hLen - 1`, where `k` is t
...
+
+A compliant implementation MUST support the following algorithms for the KEMRecipientInfo KDF and Wrap when the corresponding CompositeML-KEM algorithm is listed in KEMRecipientInfo KEM. The KDFs listed below align with the KDF used intenally for the KEM combiner. An implementation MAY also support other key-derivation functions and other key-encryption algorithms within CMS KEMRecipientInfo.
```
```suggestion
A compliant implementation MUST support the following algorithms combinations for the KEMRecipientInfo kdf and wrap fields when the corresponding Composite ML-KEM algorithm is listed in the KEMRecipientInfo kem field. The KDFs listed below align with the KDF used internally for the KEM combiner. An implementation MAY also support other key-derivation functions and other key-encryption algorithms within CMS KEMRecipientInfo.
```

danvangeest: I'm making a bunch of theses "CompositeML-KEM" -> "Composite ML-KEM" suggestions.  I've only found "CompositeML-KEM" in the edits for this PR, not elsewhere in the file.  Maybe "CompositeML-KEM" is what you want (I don't think it should be; what's a "CompositeML"?), but either way it should be consistent throughout the document.  I think all instances of "Composite KEM" should be changed to "Composite ML-KEM".

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -782,29 +770,86 @@ Note: The mask length, according to [RFC8017], is `k - hLen - 1`, where `k` is t
...
+| id-MLKEM1024-X448                 | id-sha3-512             | id-aes256-Wrap     |
+{: #tab-cms-kdf-wrap title="Mandatory-to-implement pairings for CMS KDF and WRAP"}
 
-Note: `id-aes256-Wrap` is stronger than necessary for the MLKEM768 combinations at the NIST level 3 192 bit security level, however `id-aes256-Wrap` was chosen because it has better general adoption than `id-aes192-Wrap`.
 
 where:
 
+* `id-alg-hkdf-with-sha*` are defined in [[RFC8619]].
 * `id-sha3-*` KDF instantiations are defined in {{I-D.ietf-lamps-cms-sha3-hash}}.
 * `id-aes*-Wrap` are defined in [RFC3394].
```
```suggestion
* `id-aes*-wrap` are defined in [RFC3394].
```

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -782,29 +770,86 @@ Note: The mask length, according to [RFC8017], is `k - hLen - 1`, where `k` is t
...
+|---------                          | ---                     | ---                |
+| id-MLKEM768-RSA2048               | id-alg-hkdf-with-sha256 | id-aes128-Wrap     |
+| id-MLKEM768-RSA3072               | id-alg-hkdf-with-sha256 | id-aes128-Wrap     |
+| id-MLKEM768-RSA4096               | id-alg-hkdf-with-sha256 | id-aes128-Wrap     |
+| id-MLKEM768-X25519                | id-sha3-256             | id-aes128-Wrap     |
+| id-MLKEM768-ECDH-P384             | id-alg-hkdf-with-sha384 | id-aes256-Wrap     |
+| id-MLKEM768-ECDH-brainpoolP256r1  | id-alg-hkdf-with-sha384 | id-aes256-Wrap     |
+| id-MLKEM1024-ECDH-P384            | id-sha3-512             | id-aes256-Wrap     |
+| id-MLKEM1024-ECDH-brainpoolP384r1 | id-sha3-512             | id-aes256-Wrap     |
+| id-MLKEM1024-X448                 | id-sha3-512             | id-aes256-Wrap     |
```
```suggestion
| id-MLKEM768-RSA2048               | id-alg-hkdf-with-sha256 | id-aes128-wrap     |
| id-MLKEM768-RSA3072               | id-alg-hkdf-with-sha256 | id-aes128-wrap     |
| id-MLKEM768-RSA4096               | id-alg-hkdf-with-sha256 | id-aes128-wrap     |
| id-MLKEM768-X25519                | id-sha3-256             | id-aes128-wrap     |
| id-MLKEM768-ECDH-P384             | id-alg-hkdf-with-sha384 | id-aes256-wrap     |
| id-MLKEM768-ECDH-brainpoolP256r1  | id-alg-hkdf-with-sha384 | id-aes256-wrap     |
| id-MLKEM1024-ECDH-P384            | id-sha3-512             | id-aes256-wrap     |
| id-MLKEM1024-ECDH-brainpoolP384r1 | id-sha3-512             | id-aes256-wrap     |
| id-MLKEM1024-X448                 | id-sha3-512             | id-aes256-wrap     |
```

danvangeest: RFC 3394 doesn't capitalize the W in id-aesXXX-wrap.

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -782,29 +770,86 @@ Note: The mask length, according to [RFC8017], is `k - hLen - 1`, where `k` is t
...
+
+A compliant implementation MUST support the following algorithms for the KEMRecipientInfo KDF and Wrap when the corresponding CompositeML-KEM algorithm is listed in KEMRecipientInfo KEM. The KDFs listed below align with the KDF used intenally for the KEM combiner. An implementation MAY also support other key-derivation functions and other key-encryption algorithms within CMS KEMRecipientInfo.
 
-In all cases above, the hash function used as a KDF produces a longer output than needed by the encryption algorithm, so the output SHALL be truncated to the correct length -- ie the leftmost significant bits are used.
+| Composite KEM OID                 | KDF                     | Wrap |
+|---------                          | ---                     | ---                |
+| id-MLKEM768-RSA2048               | id-alg-hkdf-with-sha256 | id-aes128-Wrap     |
+| id-MLKEM768-RSA3072               | id-alg-hkdf-with-sha256 | id-aes128-Wrap     |
+| id-MLKEM768-RSA4096               | id-alg-hkdf-with-sha256 | id-aes128-Wrap     |
+| id-MLKEM768-X25519                | id-sha3-256             | id-aes128-Wrap     |
```
danvangeest: id-sha3-256/512 is not a KDF that can be used by KEMRecipientInfo. RFC 9629 requires the KDF to take IKM, L and info parameters.  A straight SHA3 invocation doesn't take info or length parameters.

Your options are id-alg-hkdf-with-sha3-256 or id-kmac128.

If you use KMAC, you can steal section 2.2.2 of draft-ietf-lamps-cms-kyber-04 as a parallel for your "Use of the HKDF-based Key Derivation Function" section: https://datatracker.ietf.org/doc/html/draft-ietf-lamps-cms-kyber-04#section-2.2.2

If you use HKDF with SHA3, change the "HKDF may be used with different hash functions..." paragraph of the "Use of the HKDF-based Key Derivation Function" section to mention SHA3 and HKDF with SHA3 identifiers.

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -782,29 +770,86 @@ Note: The mask length, according to [RFC8017], is `k - hLen - 1`, where `k` is t
...
+| id-MLKEM768-ECDH-P384             | id-alg-hkdf-with-sha384 | id-aes256-Wrap     |
```
danvangeest: Per table 3 of https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-57pt1r5.pdf, I don't believe that SHA-384 and SHA3-512 are necessary (although perhaps you're matching with the internal Composite ML-KEM KDF.

SHA-256/SHA3-256/KMAC128 all have >= 256 bits of (2nd) preimage resistance, which is what's needed for a KDF.  Collision resistance isn't needed , there SHA-256, etc have only 128 bits of security.

By this same logic, your KEM Combiner KDFs can also be reduced to use only SHA-256 and SHA3-256 instead of any higher variants.

danvangeest (changes requested):
Also, from Section 5:

" _HKDF_: [RFC5869]. Salt is not provided; ie the default salt (all zeroes of length HashLen) will be used. In all cases, the output length of HKDF is the same as the block size of the underlying hash function, for example `HKDF-SHA256/256` means HKDF-SHA256 with an output length `L` of 256 bits (32 octets)."

"In all cases..." is not true anymore, because the output size of HKDF when used in CMS KEMRecipientInfo depends on the key size of the wrap function.  Now you will have to be explicit about the output sizes in the instances where HKDF is used.

draft-ietf-lamps-pq-composite-kem.md
ounsworth: Good suggestion. I'll do that.

draft-ietf-lamps-pq-composite-kem.md
ounsworth: Oh rats, that's a really good point. Ok. So I can't *fully* align this with the internal hash function used within the KEM combiner. Of course it needs to have the API of a real KDF.

ounsworth: Very helpful, thanks.

draft-ietf-lamps-pq-composite-kem.md
ounsworth: Very helpful, thanks.

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -718,6 +725,8 @@ A single invocation of SHA3 is known to behave as a dual-PRF, and thus is suffic
 
 The lower security levels are provided with HKDF-SHA2 as the KDF in order to facilitate implementations that do not have easy access to SHA3 outside of the ML-KEM function. Higher security levels are paired with SHA3 for computational efficiency, and the Edwards Curve (X25519 and X448) combinations are paired with SHA3 for compatibility with other similar spicifications.
 
+While it may seem odd to use 256-bit hash functions at all security levels, this aligns with ML-KEM which produces a 256-bit shared secret key at all security levels. SHA-256, SHA3-256, KMAC128 all have >= 256 bits of (2nd) preimage resistance, which is the required property for a KDF to provide 128 bits of security, as allowed in Table 3 of {{SP.800-57pt1r5}}.
```
danvangeest: Something's wrong here and I'm not sure what you're trying to say.  Per Table 3, KMAC128 has 128 bits of preimage resistance, not >= 256.  And KMAC hasn't even been brought up yet in this document, so if you're just referring to the KDFs in #tab-kem-algs, remove the mention of KMAC and say they all provider 258 bits of security, not 128.

danvangeest: And I realize the issue, in one of my previous comments I said that KMAC-128 was one of your options.  I should have said KMAC-256, sorry.  I was confusing KMAC-128 with SHAKE128/256, which does have 256 bits of security.  But per table 3 of SP.800-57pt1r5, KMAC128 has 128 bits and KMAC256 has 256 bits.  Since you're only using ML-KEM-768 and up, KMAC256 makes sense.

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -826,7 +835,26 @@ info:
...
 
-HKDF may be used with different hash functions, including SHA-256 {{FIPS.180-4}}. The object identifier id-alg-hkdf-with-sha256 is defined in {{RFC8619}}, and specifies the use of HKDF with SHA-256. The parameter field MUST be absent when this algorithm identifier is used to specify the KDF for ML-KEM in KemRecipientInfo.
+HKDF may be used with different hash functions, including SHA-256 {{FIPS.180-4}}. The object identifier id-alg-hkdf-with-sha256 is defined in [RFC8619], and specifies the use of HKDF with SHA-256. The parameter field MUST be absent when this algorithm identifier is used to specify the KDF for ML-KEM in KemRecipientInfo.
+
+### Use of the KMAC-based Key Derivation Function
+
+KMAC128-KDF is a KMAC-based KDF specified for use in CMS in {{I-D.ietf-lamps-cms-sha3-hash}}. The definition of KMAC is copied here for convenience.  Here, KMAC# indicates the use of either KMAC128-KDF or KMAC256-KDF, although only KMAC128 is used in this specification.
```
danvangeest: Per my comment above, all this (and the table below) should be KMAC256 instead of KMAC128.

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -838,21 +866,22 @@ A compliant implementation MUST support the following algorithm combinations for
...
-| id-MLKEM768-ECDH-brainpoolP256r1  | id-alg-hkdf-with-sha384 | id-aes256-wrap     |
-| id-MLKEM1024-ECDH-P384            | id-sha3-512             | id-aes256-wrap     |
-| id-MLKEM1024-ECDH-brainpoolP384r1 | id-sha3-512             | id-aes256-wrap     |
-| id-MLKEM1024-X448                 | id-sha3-512             | id-aes256-wrap     |
+| id-MLKEM768-X25519                | id-kmac128              | id-aes128-wrap     |
+| id-MLKEM768-ECDH-P384             | id-alg-hkdf-with-sha256 | id-aes256-wrap     |
+| id-MLKEM768-ECDH-brainpoolP256r1  | id-alg-hkdf-with-sha256 | id-aes256-wrap     |
+| id-MLKEM1024-ECDH-P384            | id-kmac128              | id-aes256-wrap     |
+| id-MLKEM1024-ECDH-brainpoolP384r1 | id-kmac128              | id-aes256-wrap     |
+| id-MLKEM1024-X448                 | id-kmac128              | id-aes256-wrap     |
```
danvangeest: I can't figure out the logic in the choice between KMAC and HKDF in this table. In particular, sometimes ECDH is paired with HKDF and sometimes it's paired with KMAC.

Doesn't the same logic from cms-kyber apply here - implementations might not have SHA3 at the CMS level, so that's why we use HKDF-with-SHA2 everywhere?

I realize that in my previous review I said your options here were KMAC vs HKDF-with-SHA3, but I clearly didn't look at the pairings closely enough to realize I didn't understand why things were paired as they were.

ounsworth:
> Also, from Section 5:
> 
> " _HKDF_: [RFC5869]. Salt is not provided; ie the default salt (all zeroes of length HashLen) will be used. In all cases, the output length of HKDF is the same as the block size of the underlying hash function, for example `HKDF-SHA256/256` means HKDF-SHA256 with an output length `L` of 256 bits (32 octets)."
> 
> "In all cases..." is not true anymore, because the output size of HKDF when used in CMS KEMRecipientInfo depends on the key size of the wrap function. Now you will have to be explicit about the output sizes in the instances where HKDF is used.

You're right that I needed to clean up that sentence a bit, but your conclusion is not right. This table describes the KDF used inside Composite ML-KEM. With the most recent change, Composite ML-KEM will always output a 256-bit shared secret key (same as ML-KEM). The KDF used with CMS KEMRI (specified separately in Table #tab-cms-kdf-wrap) has a variable output size to match the wrap function.

draft-ietf-lamps-pq-composite-kem.md
ounsworth: Hi @danvangeest,

The choices have been debated to death on the LAMPS and OpenPGP mailing lists. I have tried to provide some kind of coherent explanation in the very next section "## Rationale for choices", so let me know if that doesn't make sense, but the truth is that I can't really defend the choices beyond that I'm trying to balance about 15 conflicting opinions on the mailing list, and while I'm never going to make everyone happy, this set seems to be some kind of global minimum in how much I get yelled at.


The main things I'm trying to align against are:

1. X-Wing (draft-connolly-cfrg-xwing-kem-04) uses SHA3-256, so our id-MLKEM768-X25519 should do the same.
2. openpgp uses SHA3-256 for all combinations -- https://datatracker.ietf.org/doc/html/draft-ietf-openpgp-pqc-04#name-key-combiner -- Russ has been adamant that these two drafts align as much as possible.
3. Some people want SHA2 combinations, so I somewhat arbitrarily decided to make the ML-KEM768's (except X-Wing) use SHA2, and the ML-KEM1024's use SHA3. This is supposed to be some kind of compromise instead of just 2x’ing the list to provide every combination with both SHA2 and SHA3.

I guess the test will be if anyone says that the NEED ML-KEM1024 with SHA2, in which case I'll have to ask the LAMPS mailing list whether Objective 2 should be completely thrown out the window in order to satisfy Objective 3, or if I should just 2x the list.

draft-ietf-lamps-pq-composite-kem.md
ounsworth: Oh I see my mistake.

SP.800-56pt1r5 Table 3: I didn't read carefully enough. SHA-256 and SHA3-256 are equivalent to KMAC128 when you need collision resistance, but are equivalent to KMAC256 when you only need (2nd) pre-image resistance. Thanks.

ounsworth: Done.

draft-ietf-lamps-pq-composite-kem.md
ounsworth: Done.

PR #100, opened 2025-01-05
HKDF tweaks

### LINKED ISSUE #87: HKDF-based hybrid variants are unimplementable

### LINKED ISSUE #93: Need to specify clearly how to invoke HKDF  in a FIPS-compliant way

ounsworth:
Closes #87
Closes #93

### DIFF
--- a/draft-ietf-lamps-pq-composite-kem.md
+++ b/draft-ietf-lamps-pq-composite-kem.md
@@ -1566,9 +1566,9 @@ One of the primary NIST documents which is relevant for certification of a compo
 
 [SP.800-56Cr2] section 4 "One-Step Key Derivation" requires a `counter` which begins at the 4-byte value 0x00000001. However, the counter is allowed to be omitted when the hash function is executed only once, as specified on page 159 of the FIPS 140-3 Implementation Guidance [FIPS-140-3-IG].
 
-The HKDF-SHA2 options can be certified under SP.800-56Cr2 One-Step Key Derivation Option 2: `H(x) = HMAC-hash(salt, x)` where `salt` is the fixed value `0x00000000000000000000000000000000` (ie 64 bytes of zeros) in order to satisfy the requirement in [SP.800-56Cr2] that "nn the absence of an agreed-upon alternative – the default_salt shall be an all-zero byte string whose bit length equals that specified as the bit length of an input block for the hash function, hash". Note that since the desired shared secret key output length of 256 bits for all security levels aligns with the block size of SHA256, we do not need to use the HKDF-Extract step specified in [RFC5869], which further simplifies FIPS certification by allowing us to use the One-Step KDF rather than the Two-Step KDF from [SP.800-56Cr2].
+The HKDF-SHA2 options can be certified under SP.800-56Cr2 One-Step Key Derivation Option 2: `H(x) = HMAC-hash(salt, x)` where `salt` is the fixed value `0x00..00` (ie 64 bytes of zeros for SHA2-256 and 96 bytes of zeros for SHA2-384) in order to satisfy the requirement in [SP.800-56Cr2] that "nn the absence of an agreed-upon alternative – the default_salt shall be an all-zero byte string whose bit length equals that specified as the bit length of an input block for the hash function, hash". Note that since the desired shared secret key output length of 256 bits for all security levels aligns with the block size of SHA256, we do not need to use the HKDF-Extract step specified in [RFC5869], which further simplifies FIPS certification by allowing us to use the One-Step KDF rather than the Two-Step KDF from [SP.800-56Cr2].
 
-The SHA3 options can be certified under [SP.800-56Cr2] One-Step Key Derivation Option 2: `H(x) = HMAC-hash(salt, x)` with the salt omitted.
+The SHA3 options can be certified under [SP.800-56Cr2] One-Step Key Derivation Option 1: `H(x) = hash(x)`.
 
 
 ## Backwards Compatibility {#sec-backwards-compat}
--- a/draft-ietf-lamps-pq-composite-kem.md
+++ b/draft-ietf-lamps-pq-composite-kem.md
@@ -557,8 +557,8 @@ Encap Process:
         ss = SHA3-256(mlkemSS || tradSS || tradCT || tradPK || Domain)
       else if KDF is "HKDF"
         ss = HKDF-Extract(salt="", IKM=mlkemSS || tradSS || tradCT || tradPK || Domain)
-          # Note: salt is the empty string (0 octets), which will internally be mapped 
-          # to the zero vector `0x00..00` of the correct input size for the underlying 
+          # Note: salt is the empty string (0 octets), which will internally be mapped
+          # to the zero vector `0x00..00` of the correct input size for the underlying
           # hash function.
 
   6. Output composite shared secret key and ciphertext
--- a/draft-ietf-lamps-pq-composite-kem.md
+++ b/draft-ietf-lamps-pq-composite-kem.md
@@ -550,11 +550,13 @@ Encap Process:
 
   4. Encode the ciphertext
 
-     ct = mlkemCT || tradCT
+      ct = mlkemCT || tradCT
 
   5. Combine the KEM secrets and additional context to yield the composite shared secret
-
-     ss = KDF(mlkemSS || tradSS || tradCT || tradPK || Domain)
+      if KDF is "SHA3-256"
+        ss = SHA3-256(mlkemSS || tradSS || tradCT || tradPK || Domain)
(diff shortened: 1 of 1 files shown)


ounsworth:
@chris-wood I would like your review on this.

Regardless of how you feel about whether or not SHA2 combiners should be included, does this PR address the issue that the HKDF combiner was under-specified?

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -1561,11 +1564,11 @@ TODO: update this to NIST SP 800-227, once it is published.
...
 
-The HKDF-SHA2 options can be certified under SP.800-56Cr2 One-Step Key Derivation Option 1: `H(x) = hash(x)`.
+The HKDF-SHA2 options can be certified under SP.800-56Cr2 One-Step Key Derivation Option 2: `H(x) = HMAC-hash(salt, x)` where `salt` is the fixed value `0x0000000000000000` (ie 32 bytes of zeros) in order to satisfy the requirement in [SP.800-56Cr2] that "nn the absence of an agreed-upon alternative – the default_salt shall be an all-zero byte string whose bit length equals that specified as the bit length of an input block for the hash function, hash". Note that since the desired shared secret key output length of 256 bits for all security levels aligns with the block size of SHA256, we do not need to use the HKDF-Extract step specified in [RFC5869], which further simplifies FIPS certification by allowing us to use the One-Step KDF rather than the Two-Step KDF from [SP.800-56Cr2].
```
ilaril: The input block length of SHA-256 is 64 bytes, not 32, so the salt should presumably be 64 bytes of zeros, not 32. However, both will lead to the same output (due to how HMAC works).

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -550,11 +550,13 @@ Encap Process:
...
-     ct = mlkemCT || tradCT
+      ct = mlkemCT || tradCT
 
   5. Combine the KEM secrets and additional context to yield the composite shared secret
-
-     ss = KDF(mlkemSS || tradSS || tradCT || tradPK || Domain)
+      if KDF is "SHA3-256"
+        ss = SHA3-256(mlkemSS || tradSS || tradCT || tradPK || Domain)
+      else if KDF is "HKDF-SHA256"
+        ss = HKDF-SHA256(salt=0x0000000000000000, IKM=mlkemSS || tradSS || tradCT || tradPK || Domain)
```
ilaril: Is that clear enough from parameter names, or should that explicitly be HKDF-Extract-SHA256?

There are two HKDF operations (HKDF-Extract and HKDF-Expand), and parameters being salt and IKM only makes sense for HKDF-Extract.

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -1561,11 +1564,11 @@ TODO: update this to NIST SP 800-227, once it is published.
...
 
-The HKDF-SHA2 options can be certified under SP.800-56Cr2 One-Step Key Derivation Option 1: `H(x) = hash(x)`.
+The HKDF-SHA2 options can be certified under SP.800-56Cr2 One-Step Key Derivation Option 2: `H(x) = HMAC-hash(salt, x)` where `salt` is the fixed value `0x0000000000000000` (ie 32 bytes of zeros) in order to satisfy the requirement in [SP.800-56Cr2] that "nn the absence of an agreed-upon alternative – the default_salt shall be an all-zero byte string whose bit length equals that specified as the bit length of an input block for the hash function, hash". Note that since the desired shared secret key output length of 256 bits for all security levels aligns with the block size of SHA256, we do not need to use the HKDF-Extract step specified in [RFC5869], which further simplifies FIPS certification by allowing us to use the One-Step KDF rather than the Two-Step KDF from [SP.800-56Cr2].
 
-The SHA3 options can be certified under SP.800-56Cr2 One-Step Key Derivation Option 2: `H(x) = HMAC-hash(salt, x)` with the salt omitted.
+The SHA3 options can be certified under [SP.800-56Cr2] One-Step Key Derivation Option 2: `H(x) = HMAC-hash(salt, x)` with the salt omitted.
```
ilaril: The SHA-3 one seems to be just SHA3-256, which would be option 1 (hash(x)), not option 2?

draft-ietf-lamps-pq-composite-kem.md
ounsworth: Good catch. Thank you.

draft-ietf-lamps-pq-composite-kem.md
ounsworth: I've been debating with myself whether this should be `HKDF-Extract()`, or `HKDF-Expand( HDKF-Extract() )`.

In the end, I think there is no need for stretching, so only HKDF-Extract() is needed, and that makes things simpler.

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -1566,9 +1566,9 @@ One of the primary NIST documents which is relevant for certification of a compo
...
+The HKDF-SHA2 options can be certified under SP.800-56Cr2 One-Step Key Derivation Option 2: `H(x) = HMAC-hash(salt, x)` where `salt` is the fixed value `0x00..00` (ie 64 bytes of zeros for SHA2-256 and 96 bytes of zeros for SHA2-384) in order to satisfy the requirement in [SP.800-56Cr2] that "nn the absence of an agreed-upon alternative – the default_salt shall be an all-zero byte string whose bit length equals that specified as the bit length of an input block for the hash function, hash". Note that since the desired shared secret key output length of 256 bits for all security levels aligns with the block size of SHA256, we do not need to use the HKDF-Extract step specified in [RFC5869], which further simplifies FIPS certification by allowing us to use the One-Step KDF rather than the Two-Step KDF from [SP.800-56Cr2].
```
ilaril: Nit: Input block size of SHA384 is 128 bytes (16 words of 8 bytes each).

(Not that it makes actual difference as HMAC would zero-extend the key anyway.)

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -1581,7 +1584,7 @@ These migration and interoperability concerns need to be thought about in the co
...
+With regard to the traditional algorithms, RSA or Ellipti c Curve, in order to achieve the public-key binding property the KEM combiner used to form the Composite ML-KEM, the combiner requires the traditional public key as input to the KDF that derives the output shared secret. Therefore it is required to carry the public key within the respective `OneAsymmetricKey.publicKey` as per the private key encoding given in {{sec-priv-key}}. Implementers who choose to use a different private key encoding than the one specified in this document MUST consider how to provide the component public keys to the decapsulate routine. While some implementations might contain routines to computationally derive the public key from the private key, it is not guaranteed that all implementations will support this; for this reason the interoperable composite private key format given in this document in {{sec-priv-key}} requires the public key of the traditional component to be included.
```
janklaussner: this is a typo

draft-ietf-lamps-pq-composite-kem.md
ounsworth: Right. I need to actually go read FIPS 180-4 so that I don't continue getting this wrong.

ounsworth: Wait. HKDF (RFC 5869) says:
> if not provided, it is set to a string of HashLen zeros

And some of its test vectors have `salt = (0 octets)`, so maybe it's easier for me to do the same instead of trying to figure out the correct-width input? As you say, HKDF will internally handle this properly.

ounsworth: @ilaril Can you please re-check that the current wording is ok?

ounsworth:
I'm going to merge this so that we can proceed with the other PRs that depend on it. If Ilari has more comments, we can address those separately.

PR #101, opened 2025-01-06
Added appdx comparing with X-Wing and ETSI CatKDF.

### LINKED ISSUE #91: Composite KEM does not fully protect against implementation errors in ML-KEM
See comment from Peter C:

https://mailarchive.ietf.org/arch/msg/spasm/Fq06ghviNMejesJNR4XzP5x0ecI/

ounsworth:
Closes #91

TODO before merging:

- [ ] Review from Peter C, especially:
  - the new appendix comparing this specification with ETSI CatKDF
  - The intro is still a bit misleading about the endorsement from BSI, which Peter C correctly points out is to an earlier version of the composite draft, however we have Jan Klaußner as an author and close collaboration with Stavros Kousidis on the CFRG version, so BSI is a bull collaborator on developing this.
- [ ] Review from Chris Wood / Deirdre / Bas, especially:
  - The new appendix comparing this specification with X-Wing
- [ ] I was not previously aware of ETSI TS 103 744's CatKDF. We should have a discussion on the LAMPS list about whether we want to add the ML-KEM CT to the KDF in order to offer equivalent security properties as ETSI CatKDF.


draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -1585,6 +1591,35 @@ With regard to the traditional algorithms, RSA or Elliptic Curve, in order to ac
 
 <!-- End of Implementation Considerations section -->
 
+# Comparison with other Hybred KEMs
+
+## X-Wing
+
+Thit specification borrows extensively from the analysis and KEM combiner construction presented in [X-Wing]. In particular, X-Wing and id-MLKEM768-X25519 are largely interchangeable. The one difference is that X-Wing uses a combined KeyGen function to generate the two component private keys from the same seed, which gives some additional binding properies. However, using a derived value as the seed for ML-KEM.KeyGen_internal() is explicitely disallowed by [FIPS.203] which makes it impossible to create a FIPS-compliant implentation of X-Wing. For this reason, this specification keeps the key generatation for both components separate so that implementers are free to use an existing certified hardware or software module for one or both components.
```
bwesterb: - Typo: *Thit*
- "is explicitely disallowed". Presently.
- "it impossible to create a FIPS-compliant" of X-Wing private key import.

ounsworth: Thanks, fixed.

Yeah, if NIST eventually softens the prohibition on derived seeds, that would be fantastic and then X-Wing and Composite could fully collapse together, but NIST has not given any indication of when / how / if they are going to do that, so for now I think we need to proceed like this. Sigh.

bwesterb: Fantastic indeed: https://groups.google.com/a/list.nist.gov/g/pqc-forum/c/lMTIIPu9yDY/m/38felspKBQAJ !

bwesterb: Given the stated plan, and the fact that there are no ML-KEM-768 modules yet, wouldn't it make sense for the long term to align id-MLKEM768-X25519  with X-Wing. Of course you would want to do something different for the P256 one.

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -1585,6 +1591,35 @@ With regard to the traditional algorithms, RSA or Elliptic Curve, in order to ac
...
+4) Return key_material.
+~~~
+
+While not binary compatible with the KEM combiner presented in this specification, it is largely security-equivalent except for the following points.
+
+This specification binds the public key of the traditional component into the KDF.
+
+This specification does not bind the ML-KEM ciphertext -- which is MA in the ETSI CatKDF.
+
+EDNOTE / TODO: we should discuss on the LAMPS list whether to add this back in.
```
ounsworth:
```suggestion

```

Remove this ednote.

ounsworth:
I should do another pass at making the text clearer. "This specification" is unclear -- I should say "Composite KEM".
Should also describe more clearly that that paragraph is just highlighting that these are different constructions.

draft-ietf-lamps-pq-composite-kem.md
ounsworth: While I agree in principle, unfortunately a plan isn't enough. There are ML-KEM modules going through FIPS as we speak that do not even support SEED, let alone seed-from-kdf, and there are claims that seed-based private keys in general are not allowed in a CMVP-validated module. So unfortunately, I think Composites has to stay the way it is and support what is currently allowed in FIPS CMVP.

PR #103, opened 2025-01-26
Adjusted the combiner of id-MLKEM1024-ECDH-P384 from SHA3 to SHA2-384

### LINKED ISSUE #102: No composite is currently compatible with CNSA 2.0
It just came to my attention (thanks Uri), that CNSA 2.0 allows SHA3 internally to ML-DSA / ML-KEM, but not as a standalone hash function (only SHA2-384 and SHA2-512). It also allows only the Level V parameter sets for ML-DSA and ML-KEM.

That means that we currently have zero CompositeML-KEM variants that are compatible with CNSA 2.0 because we use SHA3 as the KEM Combiner on the Level V’s.

https://lamps-wg.github.io/draft-composite-kem/draft-ietf-lamps-pq-composite-kem.html#name-algorithm-identifiers

I assume this is a problem we should fix by changing id-MLKEM1024-ECDH-P384 to use HKDF-SHA2-512 ?

NSA’s stance on hybrids is well known, but I think it’s still preferrable that these exist at least one CompositeKEM that is compatible with CNSA 2.0, right?

ounsworth:
… so that it is compliant with CNSA 2.0.

Closes #102 

Rebased on top of #100. Merge that one first.


ounsworth:
Not ready, more work required.

Now that the general HKDF tweaks are merged to main, I should just cherry-pick the commit that changes MLKEM1024 from SHA3 to HKDF-SHA384, and start a new PR.

ounsworth:
Replaced by #107

Issue #112, opened 2025-03-03
Add length encoding to handle variable length keys


johngray-dev:
The private key format for ML_KEM can be a seed, and expanded format or both.  The Composite KEM draft at the moment only supports the seed format.   To better align with the ML-KEM drafts at IETF, we will add a 4-byte length tag to encode the length of the ML-KEM kem, so that it can be separated.

johngray-dev:
This was resolved in #113

PR #113, opened 2025-03-03
Added a fixed 4-byte length encoding value for the first mlkem compon…

johngray-dev:
…ent so the keys and ciphertexts can be separated.


johngray-dev:
closes #112

PR #136, opened 2025-04-27
Remove DER Encoding Rules section

### LINKED ISSUE #117: Clean up "DER encoding" sentences
We have instances of "DER encoding" in the document. Many of them do not make sense anymore since we removed the ASN.1 wrapper around public keys and signatures.

Go through each one and see if it still makes sense.

Also, the entire section "Encoding Rules" probably does not need to exist anymore since we now explicitly lay out the byte encoding of the structure.

ounsworth:
Since we are not encoding as raw bytes, no need for the section on DER encoding rules.

Closes #117


draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -338,9 +337,6 @@ In addition, the following terms are used in this document:
   A combiner specifies how multiple shared secrets are combined into
   a single shared secret.
 
-**DER:**
```
janklaussner, line 341: Same commant as in -sigs: DER is used in draft, why remove it here?

janklaussner (changes requested):
The term DER is used throughout the document, so removing the reference seems not right.

ounsworth:
I am going to close this PR and try again in a fresh one.

PR #140, opened 2025-05-07
Merges the two NIST certification sections.

### LINKED ISSUE #125: We currently have two FIPS Certification sections --> Merge them!
D.1. FIPS Certification

D.1.1. FIPS certification of Combiner Function

Also, referencing a specific page number is a bad idea because IGs get updated and page numbers change.

> [SP.800-56Cr2] section 4 "One-Step Key Derivation" requires a counter
   which begins at the 4-byte value 0x00000001.  However, the counter is
   allowed to be omitted when the hash function is executed only once,
   as specified on page 159 of the FIPS 140-3 Implementation Guidance
   [FIPS-140-3-IG].

janklaussner:


janklaussner:
The original text said that we do not use the "HKDF-Extract" step, which contradicts our HKDF description:
https://github.com/lamps-wg/draft-composite-kem/blob/cc427e13f34b9ba9bcf28adf3f8955395ce53f68/draft-ietf-lamps-pq-composite-kem.md?plain=1#L583

I fixed it to "we do not need to use the HKDF-Expand step" and hope this is correct...

ounsworth (approved):

PR #143, opened 2025-05-17
Removed all the DER wrapping text

### LINKED ISSUE #114: Add reference to component algorithm section in serialization methods to make it clear how the individual component keys should be encoded.
Add reference to component algorithm section in serialization methods to make it clear how the individual component keys should be encoded.

### LINKED ISSUE #117: Clean up "DER encoding" sentences

### LINKED ISSUE #142: BIT STRING or BIT STRING{ BIT STRING }
(copy from -sigs ... the analogous thing is needed in -kems)

Stephan Müller points out that if you just read the definitions of 

```
CompositeMLDSAPublicKey ::= BIT STRING

CompositeSignatureValue ::= BIT STRING
```

in isolation, then you might assume that to place this inside a `SubjectPublicKeyInfo.subjectPublicKey` or `Certificate.signature`, you need to double-wrap this as BIT STRING{ BIT STRING }.

This is not the case and is explained in Section 5.3 Encoding Rules, but perhaps the section that defines CompositeMLDSAPublicKey and CompositeSignatureValue needs a reference to 5.3?

ounsworth:
Sister PR in Sigs:

* https://github.com/lamps-wg/draft-composite-sigs/pull/164

Removed duplicate Serialization section.

Significantly cut down the sections that deal with DER encoding and ASN.1 structures ... I think we don't need any of that now that we have explicit non-DER serialization routines defined.

Closes #114 
Closes #117 
Closes #142 

TODO before merging:

- [x] Make sure the modified ASN.1 module compiles and matches the ASN.1 in the body.


janklaussner (commented):
I fixed some smaller spelling and ASN.1 errors, now it looks sound for me.
I wonder though if we could get rid of all the individual pk-* definitions for each combination. In the end we only pass through the object identifier id and fix the key usage. But right now I can not wrap my head around how exactly it would look like....

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -1043,69 +931,67 @@ Deserialization Process:
...
-The wire encoding of a Composite ML-KEM public key is:
 
-~~~ ASN.1
-CompositeKEMPublicKey ::= BIT STRING
-~~~
-{: artwork-name="CompositeKEMPublicKey-asn.1-structures"}
+## Encoding to DER {#sec-encoding-to-der}
 
-Since RSA and ECDH component public keys are themselves in a DER encoding, the following show the internal structure of the various public key types used in this specification:
+The serialization routines presented in {{sec-serialization}} produce raw binary values. When these values are required to be carried within a DER-encoded message format such as an X.509's `subjectPublicKey BIT STRING` [RFC5280] or a CMS `KEMRecipientInfo.kemct OCTET STRING` [RFC9629], then the composite value MUST be wrapped into a DER BIT STRING or OCTET STRING in the obvious ways:
```
johngray-dev, line 943: Should we remove the CMS comment (Same comment as I had for signatures).

johngray-dev (approved):

PR #144, opened 2025-05-20
Added section on subset profile

### LINKED ISSUE #123: Add Implementation Consideration: "Pick a small number to implement"
Add a new section in Implementation Considerations that it is NOT RECOMMENDED for anyone to implement all options.

Proposed text:

># Implementation Considerations
>## Profiling down the number of options
>
>One immediately daunting aspect of this specification is the number of composite algorithm combinations.
Each option has been specified because there is a community that has a direct application for it; typically because the traditional component is already deployed in a change-managed environment, or because that specific traditional component is required for regulatory reasons.
>
>However, this large number of combinations leads either to fracturing of the ecosystem into non-interoperable sub-groups when different communities choose non-overlapping subsets to support, or on the other hand it leads to spreading development resources too thin when trying to support all options.
>
>This specification does not list any particular composite algorithm as mandatory-to-implement, however organizations that operate within specific application domains are encouraged to define profiles that select a small number of composites appropriate for that application domain.
For applications that do not have any regulatory requirements or legacy implementations to consider, it is RECOMMENDED to implement:
> (Signatures)
> * id-MLDSA65-ECDSA-P256
>
> (KEMs)
> * id-MLKEM768-X25519-SHA3-256
> * id-MLKEM768-ECDH-P256-HKDF-SHA256

ounsworth:
Closes #123


janklaussner (approved):
Looks good

PR #154, opened 2025-05-29
Clarified notation around KDFs

### LINKED ISSUE #153: PiotrPopis comments on KDF
In #129 @PiotrPopis submitted the following. I am splitting this into its own Issue for easier tracking.

@ounsworth @janklaussner
I have a problem with the provisions concerning KEM Combiner and I ask the authors for a comment or possible changes.

As for chapter D.1.1, in my opinion it is OK, in the sense that it contains provisions consistent with SP-800-227ipd.

At the same time, equation 13 of 227 is repeated in D.1.1, but in chapter 4.4 (and thus also in 4.2 and 4.3) the "Expand" function is omitted when KDF is HKDF.

In chapter 4.4 it is mentioned that many cryptographic libraries provide only a combined interface for HKDF and do not expose HKDF-Extract() and HKDF-Expand() separately, but this remark should not mean that HKDF-Expand() can be omitted at all?

In chapter 4.4 there is a mention of the info parameter, which (I think) according to version 07 should be a zero-length string according to chapter 2.3 of RFC 5869 (the wording "empty").

In my opinion, in chapter 4 the notation consistent with equation 13 should be used and the values ​​of the info parameter should be specified. Am I right?

### LINKED ISSUE #160: Add a "Implicit inputs mapped from <OID>"
and change the algorithm defns to

    Composite-ML-KEM<OID>

Relates to https://github.com/lamps-wg/draft-composite-sigs/issues/185

ounsworth:
Closes #153 
Closes #160


PiotrPopis:
@ounsworth 
1. The latest fix for KDF=HKDF clarifies a number of concerns raised in issue #153, ​​i.e. the sentence: 
_Within this specification, "HKDF" means only the "HKDF-Extract" step; the "HKDF-Expand" step is always omitted when computing a composite KEM combiner._

2.  My previous comments were rather aimed at other changes, namely to clarify that HKDF should have the following form and parameters:
(...)
else if KDF is "HKDF"
ss = Expand(HKDF-Extract(salt="", IKM=mlkemSS || tradSS || tradCT
|| tradPK || Domain), info, L)
- Note: salt is the empty string (0 octets), which will
- internally be mapped to the zero vector `0x00..00` of
- the correct input size for the underlying hash function
- as per section 2.2 of [RFC5869];
- info is a zero-length string and
- L is 32 (256 bits).
which would be consistent with the provisions of section D.1.1:
(...) _The Composite KEM variants that use SHA3 as a combiner fit form (12) while the variants that use HKDF-SHA2 fit form (13)._
Now I see that my previous comment was not understood - I should probably point out the above sentence from D.1.1.

3. What now?
The sentence from D.1.1. must be changed, but I recommend to maintain compliance with NIST SP 800-227. Since NIST approves both one-step key derivation and two-step key derivation, let's not modify the "two-step" version, which will be considered as non-compliance, and let's introduce only the "one-step" version with three hash functions: SHA256, SHA3-256 and SHA384/256 to KEM Combiner.

However, such an approach, although rather compliant with NIST 227, requires modification of section 10.2.3 "SHA3 vs HKDF-SHA2", in particular the last paragraph:
_SHA3 is known to have all of the necessary dual-PRF properties [X-Wing], but SHA2 does not and therefore all SHA2-based constructions MUST use SHA2 within an HMAC construction such as HKDF-SHA2 [GHP18]._
Of course, a change in this section must also be made if the authors decide to leave only the Extract() function, and not the entire Expand(Extract()).

My recommendation:
- NIST 227 compliance is a MUST;
- "dual-PRF properies" is of lesser importance to me, as long as NIST accepts it.

ounsworth:
Hi @PiotrPopis 

I still do not understand your comment.
We are never using the "two-step" construction from SP 56Cr2 / 227. We are only using "One-step"; we either use it in the form

    SHA3(x)

which is SP 800-56Cr2 section 4.1 Option 1, and (12) in SP 800-227.

Or else we are doing 

    HKDF-Expand(salt, x) = HMAC(salt, x)

which is SP 800-56Cr2 section 4.1 Option 2, and still (12) in SP 800-227. **This is still a one-step KDF as per NIST**.

I though this was already clearly stated in D.1.1:

> The HKDF-SHA2 options can be certified under [[SP.800-56Cr2](https://lamps-wg.github.io/draft-composite-kem/draft-ietf-lamps-pq-composite-kem.html#SP.800-56Cr2)] One-Step Key Derivation Option 2 ...

... Oh I see the problem. That paragraph is in conflict with a sentence higher up in D.1.1:

> The Composite KEM variants that use SHA3 as a combiner fit form (12) while the variants that use HKDF-SHA2 fit form (13).

Ok, yes, that sentence is incorrect. there is definitely some clean-up work to do here on D.1.1. Probably I should delete formula (13) since in the end we don't use it.

johngray-dev (approved):
Seems very clear to me.

PiotrPopis:
@ounsworth 
It is now clear to me that KEM Combiner should always use one-step key derivation according to equation (12) of NIST 227.
In the case of SHA3 it will be option 1 according to chapter 4.1 of NIST SP 800-56C rev. 2, while in the case of SHA2 it will be option 2 according to 800-56Cr2.

We agreed that the previous provisions were ambiguous in this matter, and I was additionally confused by the reference to HKDF compliant with RFC5869.

For me (and probably many others reading the document) HKDF is a function described in RFC 5859. Therefore, when I see the phrase _if KDF is "HKDF"_, I expect the familiar function from RFC 5859. Instead, I read that it is just Extract(), which is basically an HMAC with a fixed key of zeros.

I suggest - in addition to changes in D.1.1 - abandoning the reference to HKDF, and instead write directly that in the case of SHA2 we use the HMAC function (FIPS198, RFC 2104) as KDF with the key K being a string of HashLen zeros.

I confirm that the document indicates that it is only "partial" HKDF (limited only to the Extract() function), but the reader still has doubts about the "full" version of HKDF - see e.g. section 7.1, where there is a clear suggestion of "full" HKDF: _For the use of HKDF [RFC5869]: a salt is not provided; i.e. the default salt (all zeroes of length HashLen) will be used. For HKDF-SHA256 the output of 256 bit output is used directly; for HKDF-SHA384/256, HKDF is invoked with SHA384 and then the output is truncated to 256 bits, meaning that only the first 256 bits of output are used._

However, if the authors decided to leave the current provisions regarding HKDF, I recommend clarifying the following in chapter 7.1:
For the use of HKDF [RFC5869] (**reduced to just the Extract() function**): a salt is not provided; i.e. the default salt (all zeroes of length HashLen) will be used. For HKDF-SHA256 the output of 256 bit output is used directly; for HKDF-SHA384/256, HKDF is invoked with SHA384 and then the output is truncated to 256 bits, meaning that only the first 256 bits of output are used.

Additionally: I withdraw my comment regarding the modification of section 10.2.3.

ounsworth:
Closes #160

carl-wallace (approved):

johngray-dev (approved):

johngray-dev (approved):

johngray-dev:
I think this also Closes #158

PR #159, opened 2025-06-02
Changed SHA384 to SHA512

### LINKED ISSUE #158: HKDF-SHA384 vs SHA512
Richard Kettlewell pointed out:

> Mike Ounsworth John Gray I notice in https://github.com/lamps-wg/draft-composite-kem/blob/main/draft-ietf-lamps-pq-composite-kem.md that you use HKDF-SHA384-Extract truncated to 256 bits - given SHA384 is secretly a truncated SHA512 anyway is there a reason you chose that rather than HKDF-SHA512-Extract with truncated output? Not really asking for any changes, just wondering about motivation.

- [x] Make sure Richard is in the acknowledgements of both drafts

ounsworth:
Closes #158


draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -1090,15 +1091,15 @@ EDNOTE: these are prototyping OIDs to be replaced by IANA.
 | id-MLKEM768-ECDH-P256-HKDF-SHA256  | &lt;CompKEM&gt;.54   | MLKEM768        | ECDH-P256            | HKDF-SHA256 |
 | id-MLKEM768-ECDH-P384-HKDF-SHA256  | &lt;CompKEM&gt;.55   | MLKEM768        | ECDH-P384            | HKDF-SHA256 |
 | id-MLKEM768-ECDH-brainpoolP256r1-HKDF-SHA256   | &lt;CompKEM&gt;.56   | MLKEM768        | ECDH-brainpoolp256r1 | HKDF-SHA256 |
-| id-MLKEM1024-ECDH-P384-HKDF-SHA384 | &lt;CompKEM&gt;.57   | MLKEM1024       | ECDH-P384            | HKDF-SHA384/256 |
-| id-MLKEM1024-ECDH-brainpoolP384r1-HKDF-SHA384  | &lt;CompKEM&gt;.58   | MLKEM1024       | ECDH-brainpoolP384r1 | SHA3-256 |
+| id-MLKEM1024-ECDH-P384-HKDF-SHA512 | &lt;CompKEM&gt;.57   | MLKEM1024       | ECDH-P384            | HKDF-SHA512/256 |
+| id-MLKEM1024-ECDH-brainpoolP384r1-HKDF-SHA512  | &lt;CompKEM&gt;.58   | MLKEM1024       | ECDH-brainpoolP384r1 | SHA3-256 |
```
johngray-dev: Is this correct.  Should it be HKDF-SHA512 instead of SHA3-256 in the KDF column?

johngray-dev: This Paragraph:

> The lower security levels (i.e. ML-KEM768) are provided with HKDF-SHA2 as the KDF in order to facilitate implementations that do not have easy access to SHA3 outside of the ML-KEM function. Higher security levels (i.e. ML-KEM1024) are paired with SHA3 for computational efficiency except for one variant paired with HKDF-SHA384 in order to leave open the possibility of compliance with [CNSA2.0], and the Edwards Curve (X25519 and X448) combinations are paired with SHA3 for compatibility with other similar specifications.

Seems to indicate This one:  id-MLKEM1024-ECDH-P384-HKDF-SHA384  but it was changed to id-MLKEM1024-ECDH-P384-HKDF-SHA512 by this change.   

I'm still confused by the last line, the name.  If KDF is indeed SHA3-256 for line 1095

| id-MLKEM1024-ECDH-brainpoolP384r1-HKDF-SHA512  | &lt;CompKEM&gt;.58   | MLKEM1024       | ECDH-brainpoolP384r1 | SHA3-256 |

Shouldn't it be called:  id-MLKEM1024-ECDH-brainpoolP384r1-SHA3-256  ?

PiotrPopis: I understood Richard Kettlewell's comment to mean that only the KDF column in table 2 (section 7.1) needs to be corrected, where HKDF-SHA384/256 should be replaced by HKDF-SHA512/256. This is de facto an editorial change, but it introduces compatibility of hash function names with FIPS 180-4, especially sections 6.5 and 6.7.

optnfast: I think we still need some notation clarity.

My understanding (and hope l-) was that the goal was HKDF-Extract-SHA512, truncated to 256 bits. Since HDKF-Extract is just HMAC with a fixed key, we might write this:

`  (HMAC-SHA-512(0,Z))/256`

The current notation (HKDF-SHA512/256) however risks confusion with using SHA512/256 (FIPS 180-4 s5.3.6 & s6.7) as the hash. We might write this:

`  HMAC-(SHA-512/256)(0,Z)`

This is _not the same thing_: with SHA512/256 the inner hash of the HMAC is truncated, and both hashes use different initial hash values.

It could also risk confusion with the FIPS 180-4 s7 permission to just truncate hashes without changing IHVs etc.

Other notes:

* SHA512/256 is referenced by SP800-56Cr2, so I think it would still be acceptable in a FIPS environment
* I suspect SHA512/256 is not commonly implemented. It's an easy lift but really, who wants to?
* SHA384 has different initial hash values again.

ounsworth: Good point. I will remove the "HMAC-SHA512/256" notation and instead put "HMAC-SHA512" in the table, and also add an explicit "truncate(ss, 256)" to the KemCombiner function definition.

ounsworth:
I am changing the merge target: I am merging into the branch for #154 rather than Main.

PR #170, opened 2025-06-16
Fixes and alignments

janklaussner:
* refer to all EC algorithms as DH, because ECDH in most case excludes X25519 and X448
* adds secp256r1 to fixed algorithms identifiers

### DIFF
--- a/draft-ietf-lamps-pq-composite-kem.md
+++ b/draft-ietf-lamps-pq-composite-kem.md
@@ -389,7 +389,7 @@ Discussion of the specific choices of algorithm pairings can be found in {{sec-r
 
 # Overview of the Composite ML-KEM Scheme {#sec-kems}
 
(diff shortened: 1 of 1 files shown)


janklaussner:
One thing I like to crosscheck: 
https://github.com/lamps-wg/draft-composite-kem/blob/b9af933c93148f0014a329f3b514904433bb1c2a/draft-ietf-lamps-pq-composite-kem.md?plain=1#L694
I removed the tuple return in this PR because I am not aware that TradKEM.Decap() returns something else than tradSS.

johngray-dev:
Thanks for these changes Jan.  They look good.  I agree we shouldn't have to return a tuple from the decaps function when it returns a single ss.  That may have been left over from our attempt to make things more generic a few weeks ago.

johngray-dev (approved):

ounsworth (approved):
Changes look good. Thanks for the careful review.

PR #178, opened 2025-06-19
Change the encoding of the ECDSA and RSA components to PKCS1

### LINKED ISSUE #175: Incorrect private key test vectors
The test vectors are incorrect for the EC and RSA components of the composite private keys.  The private keys are encoded as a concatenation of the raw (ML-KEM,ML-DSA) seed and a PrivateKeyInfo.  Rather than a PrivateKeyInfo it should be an ECPrivateKey or RSAPrivateKey.  The 25519/448 private key vectors are correct according to the current text.

johngray-dev:
Closes #175 

Corrected the composite component encoding format.


danvangeest (approved):
Verified that the ECDSA test vectors work.  I don't have the RSS ones implemented yet, but the ASN.1 looks good.

johngray-dev:
Thanks Dan for Verifying so quickly!

PR #183, opened 2025-07-08
Changes representation of X25519/X448 private keys to CurvePrivateKey.

### LINKED ISSUE #174: x25519/x448 encoding
For the 25519/448 composites, the traditional private key component is the raw 32 or 56 bytes.  Since the EC and RSA private key components are to be encoded as ECPrivateKey and RSAPrivateKey, it might be more consistent to encode the 25519/448 keys as a CurvePrivateKey (i.e. wrap the raw key in an OCTET STRING).  I'm not convinced that doing this one way or the other is better, just that doing so would be consistent. Then all traditional components will be equivalent to the contents of a standalone traditional PrivateKeyInfo.privateKey OCTET STRING.

janklaussner:
Closes #174


danvangeest (approved):
Text looks good, I haven't reviewed the test vectors.

johngray-dev (approved):
Thanks for the Change.  Looks good Jan!

PR #189, opened 2025-07-19
write an artifact zip

xyakimo1:


ounsworth (approved):

ounsworth:
Thank you Stepan, this is great!

I'll merge this.

The next step is to make  PR to https://github.com/IETF-Hackathon/pqc-certificates  .. to add this zip. It probably makes sense to add a new folder `providers/composite-kem-ref-impl` (and maybe rename the existing `composite-ref-impl` to `composite-sigs-ref-impl`)

PR #194, opened 2025-08-16
Properly wrapped the non-composite ML-KEM test vector priv keys in an…

### LINKED ISSUE #172: Invalid encoding for non-composite ML-KEM test vectors
While the non-composite ML-KEM test vectors don't actually test composites, they're useful for verifying one's framework when using these test vectors.

The PKCS#8 non-composite ML-KEM test vectors are encoded incorrectly. They contain the seed in the private key OCTET STRING without any additional wrapping.  Because the CHOICE encoding is now used for ML-KEM, when encoded as a seed the private key needs an extra layer of OCTET STRING.

Not sure how you then want to handle the raw key.  Should it be the raw SEED, or the raw CHOICE (no PKCS#8, but still CHOICE ANS.1 encoding).

ounsworth:
… OCTET STRING to satisfy the CHOICE structure. Closes #172


ounsworth:
@danvangeest  I would like your review on this. 

This PR mirrors https://github.com/lamps-wg/draft-composite-sigs/pull/219 in -sigs.

johngray-dev:
Authors agree, please merge after resolving conflict.

src/testvectors.json
```diff
@@ -1,131 +1,131 @@
...
+      "dk": "BECFBsVCSMvxfRo1GENXouQBHRmBdgNrakFyNojQZSGmdTobhEH3CtKhlgUV1azIy6Ql1LBgAKD6w6t6KT62/WzW",
+      "dk_pkcs8": "MFQCAQAwCwYJYIZIAWUDBAQCBEKAQIUGxUJIy/F9GjUYQ1ei5AEdGYF2A2tqQXI2iNBlIaZ1OhuEQfcK0qGWBRXVrMjLpCXUsGAAoPrDq3opPrb9bNY=",
```
danvangeest: `dk_pkcs8` is now a properly encoded ML-KEM PKCS#8 (in this case the seed CHOICE which is an octet string with tag [0]) with ASN.1 type 0x80.

However, `dk` is just a regular OCTET STRING (ASN.1 type 0x04) wrapping the raw private key.  This encoding is useful to no one, it should either be the CHOICE encoding (i.e. the same as `dk_pkcs8` without the PKCS#8 wrapping), or it should be the raw seed without any ANS.1 wrapping.

Since `ek` is simply a raw public key without any ASN.1 wrapping, I think `pk` should also be the raw key without any ASN.1 wrapping.  Also Appendix F says "dk the raw decapsulation private key."

Since these tests are meant to mirror the NIST ACVP KATs, should `dk` be the expanded key?  Regardless, the text "dk the raw decapsulation private key." should be changed to a more explicit "dk the raw decapsulation private key in <seed? expanded?> format." `dk_pkcs` could be any of the CHOICE options, currently seed.

Perhaps controversially, do `ek` and `dk` actually need to be included in the test vectors when we already have `x5c` and `dk_pkcs8`?  Possibly, since some implementations may only implement the algorithm and some may also implement the encoding.

src/generate_test_vectors.py
```diff
@@ -1174,7 +1174,7 @@ def formatResults(kem, caSK, ct, ss ):
 
   # for standalone ML-KEM, we need to wrap the private key in an OCTET STRING, but not when it's a composite
```
danvangeest, line 1173: Per my other comment on this review, I believe that `dk` should just be the raw bytes.  Hence the standalone KEM logic on line 1159 should be removed and the raw bytes always set for `jsonTest['dk']`.

Only when encoding the key into PKCS#8 does there need to be different logic.  Here on line 1175 the comment should be changed from

> \# for standalone ML-KEM, we need to wrap the private key in an OCTET STRING, but not when it's a composite

to

> \# for standalone ML-KEM, we need to encode the private key using ML-DSA's CHOICE encoding, but not when it's a composite

danvangeest (changes requested):

src/testvectors.json
ounsworth: Good point. My intention was for it to be the raw seed.
And that's what's done in composite-sigs:
https://github.com/lamps-wg/draft-composite-sigs/blob/main/src/testvectors.json#L8

... so that's a bug. I'll fix that.

... somewhat embarassingly, I ended up having to put in stupid wrapper code to deal with this stray OCTET STRING ... and yet I still didn't realize that this was wrong LOL

```
  def loadKeyPair(self, private_bytes: bytes) -> None:
    if len(private_bytes) == 66:
      # there's an extra OctetString wrapper
      private_bytes = private_bytes[2:]
```

Great. Let's take that out.

ounsworth: Fixed in [5ac0c98](https://github.com/lamps-wg/draft-composite-kem/pull/194/commits/5ac0c9866ddda1dc2a08211900224eaf4fd5b333)

PR #203, opened 2025-09-02
Changes size table calculation to maximum values with fixed length hint.

### LINKED ISSUE #176: Approximate key sizes
Appendix A.  Approximate Key and Signature Sizes does not seem particularly useful, because it is just approximate.  If you could list the maximum sizes that would be much more useful. I suspect some of the private key values in this table are no longer approximate because of the serialization issues I mentioned above.

janklaussner:
Closes #176


janklaussner:
merge PR #202 first! This PR relies on the changes made there...

ounsworth:
- [x] Before merging this, I will take an action to modify the script to test the actual test vectors against the calculated maximum:

* If the value is variable, then check that the test vector is <=
* If the value is not variable, then check that the test vector is ==

ounsworth (approved):
I added test code to check the generated test vectors against the expected size table.

Everything looks good:

> DEBUG: all sizes matched expected!

PR #204, opened 2025-09-02
Fixes ECDH Public Key encoding.

### LINKED ISSUE #200: EC public key is encoded as RAW, needs to be ECPoint
TODO - Code bug - EC public key is RAW (X9.62) value, not an ECPoint which is what we decided to use.

janklaussner:
OCTET STRING wrapper is added to the X9.62 encoded EC public key as
specified in RFC 5480.

Closes #200


janklaussner:
merge PR #203 first.

janklaussner:
Cleaned up the PR and added the modified Size Table not included yet. Be aware that the TradPK is included in the CompositeSK also, hence all ECDH composite SKs also increase.
@ounsworth @johngray-dev  please review again

src/kemCombiner_MLKEM768_X25519_SHA3_256.md
```diff
@@ -2,31 +2,31 @@ Example of id-MLKEM768-X25519-SHA3-256 Combiner function output.
...
 
 tradCT:
-ce884655f255ec09d4a282a52111dd57e5f9f4c343d5e8c076713433088e3324
+cb3880dff0732770333041dc3440304d83f0d75eee2d53729fa48cabfcec9c74
 
 tradPK:
-b4dd5bdbcd9a94eac11786a68b768baadc4c1942ae95eaab5168e5f4b54af876
+009e14530569feefc03d815f6c5dceb677c25184121a18902fbead6503a84f68
 
 Label:  \.//^\
```
feventura, line 16: backslash is escaping the dot.

src/kemCombiner_MLKEM768_X25519_SHA3_256.md
```diff
@@ -2,31 +2,31 @@ Example of id-MLKEM768-X25519-SHA3-256 Combiner function output.
...
 
 
 # Combined KDF Input:
 #  mlkemSS || tradSS || tradCT || tradPK || Label
 
-Combined KDF Input: 77c7848ed7cf2ec5b986754a394290eb821e54bc7f30449148
-88e77696e0cdd83e53b2e467713641acfd9bc1939310178063dc9caea9d8b13eb94884
-c360442bce884655f255ec09d4a282a52111dd57e5f9f4c343d5e8c076713433088e33
-24b4dd5bdbcd9a94eac11786a68b768baadc4c1942ae95eaab5168e5f4b54af876\.//
```
feventura, line 25: backslash is escaping the dot.

ounsworth:
On more careful reading of RFC 5480, even though `ECPoint ::= OCTET STRING`, it has wording that when you put it into a SubjectPublicKeyInfo, you put it directly into the SPKI BIT STRING _without_ the OS wrapper. We checked a publicly-trusted EC root to confirm.

So, actually, our test vectors were correct already.

Closing this PR.

PR #210, opened 2025-09-19
Tweaked the wording of the Encoding to DER section

### LINKED ISSUE #192: Encoding to DER in obvious ways
Section 6.1 says the following:
```
... the composite value MUST be wrapped into a DER BIT STRING or OCTET STRING in the obvious ways.
```
To me something like *obvious ways* seems unfitting for an RFC. This wording is especially misleading as the ASN.1 BIT STRING type was not intended to hold byte array like data, but a sequence of flags, and in examples like certificate `subjectPublicKey` fields, one explicitly needs to keep the trailing zeros (see OpenSSL's `ossl_asn1_string_set_bits_left` for example). This may be worth clarifying.

ounsworth:
Closes #192

### DIFF
--- a/draft-ietf-lamps-pq-composite-kem.md
+++ b/draft-ietf-lamps-pq-composite-kem.md
@@ -1062,13 +1062,20 @@ While composite ML-KEM keys and ciphertext values MAY be used raw, the following
 
 ## Encoding to DER {#sec-encoding-to-der}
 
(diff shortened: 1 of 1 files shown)


ounsworth:
I got confused about branches. So this PR also contains a tweak to the 65537 text.

PR #214, opened 2025-10-01
Changed HMAC-SHA2 to SHA3 in test vectors

### LINKED ISSUE #180: FIPS vs truncated HMAC
> In alignment with ML-KEM [FIPS.203], Composite KEM algorithms output a 256-bit shared secret key at all security levels, truncating is necessary as described in {{sec-kem-combiner}}.

Where the combiner is SHA3-256 or HMAC-SHA256 this is all good.

However where the combiner is HMAC-SHA512 it means we truncate the output to 256 bits.

This is a headache in the FIPS context:
* I see nothing in SP800-227ipd which permits truncation, so we can only appeal to the underlying KDF specs.
* If the combiner is interpreted as the an SP800-56Cr2 one-step KDF we could only get away with it if the result is used directly as key material (it's conceptually the same as only needing 256 bits of KDF output).
* If the combiner is interpreted as the Extract phase of an SP800-56Cr2 two-step KDF it won't work: the text is clear that the HMAC must be untruncated.

So if you want more than one AES-256 key, or two AES-128 keys, you're out of luck. I have concrete use cases for up to 1024 bits of output from a PQ/T hybrid KEM so this is going to be a problem for me. (My current use cases don't involve interop so I do have a good plan B, but I'd still like to align with the rest of the world if possible.)

I think I have some responsibility for this, having previously cheered on the use of HMAC-SHA512.

My suggestion is to remove all truncated HMAC-SHA512 and replace with untruncated HMAC-SHA256. I don't believe this will have any meaningful impact on security level.

### LINKED ISSUE #206: SHA2 vs SHA3 for the KEM combiner
This is the only outstanding item not covered in https://github.com/lamps-wg/draft-composite-sigs/issues/244

SHA2 vs SHA3 for the KEM combiner. Alignment with cms-mldsa which uses SHA2, vs alignment with cfrg-hybrid-kems which uses SHA3. MikeO is working with Richard Barnes. We need to re-survey the community on the operational requirements behind this point. It would be nice if we could just use SHA3 throughout, maybe except for the RSA composites which CFRG doesn't have.

### LINKED ISSUE #211: Roll the OIDs for -08

ounsworth:
Closes #206 
Closes #180 
Closes #211 

- [x] Needs new OIDs from the Entrust arc.  ... #211


draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -1201,7 +1174,7 @@ Label values are provided as ASCII strings, but MUST be converted into binary st
 For example:
 
 * "\.//^\" in hexadecimal is "5c2e2f2f5e5c"
-* "QSF-MLKEM768-P256-HMACSHA256" in hexadecimal is "5153462d4d4c4b454d3736382d503235362d484d4143534841323536"
+* "QSF-MLKEM768-P256-SHA3-256" in hexadecimal is "5153462d4d4c4b454d3736382d503235362d534841332d323536"
```
danvangeest, line 1192: How much is this document aligning with draft-irtf-cfrg-concrete-hybrid-kems?  Because their label is "QSF-P256-MLKEM768-SHAKE256-SHA3256".  Which is not even consistent with their KEM naming, I've raised that here: https://github.com/cfrg/draft-irtf-cfrg-concrete-hybrid-kems/issues/23

src/algParams.md
```diff
@@ -1,108 +1,108 @@
-- id-MLKEM768-RSA2048-HMAC-SHA256
-  - OID: 2.16.840.1.114027.80.5.2.50
-  - Label: "`QSF-MLKEM768-RSAOAEP2048-HMACSHA256`"
-  - Key Derivation Function (KDF): HMAC-SHA256
+- id-MLKEM768-RSA2048-SHA3-256
+  - OID: 2.16.840.1.114027.80.5.2.62
+  - Label: "`QSF-MLKEM768-RSAOAEP2048-SHA3-256`"
```
danvangeest: As above, all these labels don't align with the label convention from draft-irtf-cfrg-concrete-hybrid-kems.

I'm not confident that they won't change that convention.

ounsworth:
TODO: did I remember to remove SHA2 from the Algorithm References Appendix table?

ounsworth:
Author's meeting today: this is good. Merge it once this is all cleaned up.

draft-ietf-lamps-pq-composite-kem.md
ounsworth: Thanks @danvangeest . I'll change the "SHA3-256" to "SHA3256".

I've also yelled at them about the order of components, which is dumb.

About the SHAKE256: I know that Richard is of the opinion that how you did the keygen shouldn't matter for for encapsulating to a public key, but I think I'll leave that one alone. The LAMPS document is out of time; it needs to go now, so I think the right thing to do is to choose HPKE-friendly labels, and if they choose not to align to them, that's their choice.

PR #218, opened 2025-10-10
WGLC pass over encodings

### LINKED ISSUE #149: Lock the ref impl github link to a specific commit
As part of WGLC, the link at the top of the Test Vectors section should be locked to a specific commit, and should also probably be moved to be a proper Informative reference.

Might as well leave this until we're pretty sure that we're done tweaking the reference implementation code, which probably means leaving it fairly late in the WGLC.

### LINKED ISSUE #217: Still some underspecified encodings
>  *ML-KEM*: MUST be encoded as specified in sections 7.1, 7.2 and
      7.3 of [FIPS.203], using a 64-byte seed as the private key.

I think 7.3 is unnecessary since it defines decapsulation, without introducing any new encodings.

>    *ECDH*: public key MUST be encoded as an uncompressed ECPoint as
      specified in section 2.2 of [RFC5480].  The private key must be
      encoded as ECPrivateKey specified in [RFC5915] without
      'NamedCurve' parameter and without 'publicKey' field.

Needs to specify ciphertext encoding.  I know it'll just same something about a certina number of raw bytes or whatever, but it should still say something.

>     *X25519 and X448*: the public key MUST be encoded as per section 5
      of [RFC7748] and the private key as CurvePrivateKey specified in
      [RFC8410].

Needs to specify ciphertext encoding.  I know it'll just same something about a certina number of raw bytes or whatever, but it should still say something.

ounsworth:
Closes #217 
Closes #149


draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -801,15 +815,16 @@ Deserialization is possible because ML-KEM has fixed-length public keys, private
...
 * **RSA**: the public key MUST be encoded as RSAPublicKey with the `(n,e)` public key representation as specified in A.1.1 of [RFC8017] and the private key representation as RSAPrivateKey specified in A.1.2 of [RFC8017] with version 0 and 'otherPrimeInfos' absent. An RSA-OAEP ciphertext MUST be encoded as specified in section 7.1.1 of {{RFC8017}}
-* **ECDH**: public key MUST be encoded as an uncompressed `ECPoint` as specified in section 2.2 of [RFC5480]. The private key must be encoded as ECPrivateKey specified in [RFC5915] without 'NamedCurve' parameter and without 'publicKey' field.
-* **X25519 and X448**: the public key MUST be encoded as per section 5 of [RFC7748] and the private key as CurvePrivateKey specified in [RFC8410].
+* **ECDH**: public key MUST be encoded as an uncompressed X9.62 [X9.62–2005], including the leading byte `0x04` indicating uncompressed. This is consistent with the encoding of `ECPoint` as specified in section 2.2 of [RFC5480] when no ASN.1 OCTET STRING wrapping is present. The private key must be encoded as ECPrivateKey specified in [RFC5915] without 'NamedCurve' parameter and without 'publicKey' field.
```
ounsworth:
```suggestion
* **ECDH**: public key MUST be encoded as an uncompressed X9.62 [X9.62–2005], including the leading byte `0x04` indicating uncompressed. This is consistent with the encoding of `ECPoint` as specified in section 2.2 of [RFC5480] when no ASN.1 OCTET STRING wrapping is present. The private key MUST be encoded as ECPrivateKey specified in [RFC5915] with 'NamedCurve' parameter set to the OID of the curve, but without the 'publicKey' field.
```

PR #228, opened 2025-10-29
WGLC review

danvangeest:


draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -1189,7 +1179,7 @@ Each Composite ML-KEM algorithm has a unique Label which is used in constructing
 Label values are provided as ASCII strings, but MUST be converted into binary strings in the obvious way.
 For example:
 
-* "\.//^\" in hexadecimal is "5c2e2f2f5e5c"
+* "\\.//^\\" in hexadecimal is "5c2e2f2f5e5c"
```
danvangeest: comment that the X-wing label sucks

danvangeest:
The change to testvectors_wrapped.json is just to rewrap them with better line wrapping (trust me bro).  In particular, previously often the "tcId" value would be wrapped in the middle which made it had to actually find algorithm test vectors in the draft,

draft-ietf-lamps-pq-composite-kem.md
sophieschmieg: we could consider just using the hex here, that way the whole backslash backlash goes away

PR #243, opened 2025-11-05
WGLC2

### LINKED ISSUE #233: DVG WGLC comment: The TLA "CVE" is used but not defined

### LINKED ISSUE #234: DVG WGLC comment: remove CMS paragraph
2. Section 3: The following paragraph seems unnecessary.  We don't say anything similar for kyber-certificates.  This was probably added when the CMS parts were split into a pending draft.

>    The KEM interface defined above differs from both traditional key
   transport mechanism (for example for use with KeyTransRecipientInfo
   defined in [RFC5652]), and key agreement (for example for use with
   KeyAgreeRecipientInfo defined in [RFC5652]) and thus Composite ML-KEM
   MUST be used with KEMRecipientInfo defined in [RFC9629], however full
   conventions for use of Composite ML-KEM within the Cryptographic
   Message Syntax will be included in a separate specification.

### LINKED ISSUE #236: DVG WGLC comment: how is ciphertext not attacker controlled?
5. Section 4.3: How is ciphertext not attacker controlled?  If anything, this could say that because DER uses length-encoding these variations are benign (my PR says something similar a few lines up but I wanted to point this sentence out explicitly because of the "attacker controlled" part).

>    Implementers should keep in
   mind that some instances of tradCT and tradPK will be DER-encoded
   which could introduce minor length variations such as dropping
   leading zeroes; since these variations are not attacker-controlled
   they are considered benign

### LINKED ISSUE #237: DVG WGLC comment: Add Security Considerations reference to sfluhrer-cfrg-ml-kem-security-considerations
6. Section 10: Should there be a reference to sfluhrer-cfrg-ml-kem-security-considerations somewhere in the Security Considerations?

### LINKED ISSUE #238: DVG WGLC comment: Remove section 10.4
7. Section 10.4: Most or all other ML-KEM-related drafts throughout the IETF have stopped mentioning the probability of ML-KEM decapsulation failure because it is so vanishingly small.  I'd recommend removing this section.

### LINKED ISSUE #241: Incorrect text for the private key of X25519 and X448.
```
   *  *X25519 and X448*: the public key MUST be encoded as per section 5
      of [RFC7748] and the private key is a 32 or 57 byte raw value for
      Ed25519 and Ed448 respectively.  The ciphertext MUST be encoded in
      the same manner as the public key.
```
Shall be changed to (changes are enclosed with **)
```
   *  *X25519 and X448*: the public key MUST be encoded as per section 5
      of [RFC7748] and the private key is a 32 or 5**6** byte raw value for
      **X**25519 and **X**448 respectively.  The ciphertext MUST be encoded in
      the same manner as the public key.
```

Here the text from section of [RFC7748]:
```
   The "X25519" and "X448" functions perform scalar multiplication on
   the Montgomery form of the above curves.  (This is used when
   implementing Diffie-Hellman.)  The functions take a scalar and a
   u-coordinate as inputs and produce a u-coordinate as output.
   Although the functions work internally with integers, the inputs and
   outputs are 32-byte strings (for X25519) or 56-byte strings (for
   X448) and this specification defines their encoding.
```

ounsworth:
Closes #233
Closes #234 
Closes #236 
Closes #237 
Closes #238 
Closes #241


johngray-dev (approved):
Read through the changes,  it looks good to me.  Thanks for addressing all these comments Mike.

PR #244, opened 2025-11-06
Reverted priv key format to mlkemSeed || tradSK

ounsworth:


danvangeest (commented):
cycle OIDs?

johngray-dev:
Yes, I'll cycle them right now...

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -1508,9 +1464,62 @@ In applications that only allow NIST PQC Level 5, it is RECOMMENDED to focus imp
...
+2. Fetch it from an external data source, for example from the public-key certificate corresponding to this private key.
+
+3. If the composite KEM private key is being carried within a PKCS#8 OneAsymmetricKey object, place the full composite public key within the optional OneAsymmetricKey.publicKey field, which allows extracting the tradPK (and re-encode as necessary for correctly using it in the KEM Combiner).
+
+4. Use an alternate private key encoding that explicitly carries the tradPK.
+
+
+### Extracting RSAPublicKey from RSAPrivateKey {#sec-rsa-pub-from-priv}
+
+Assuming that the RSA component of the composite private key is encoded as a RSAPrivateKey, as required by this specification, then, quoting from [RFC8017] you have:
```
xipki: a RSAPrivateKey -> an RSAPrivateKey

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -1508,9 +1464,62 @@ In applications that only allow NIST PQC Level 5, it is RECOMMENDED to focus imp
...
+RSAPublicKey ::= SEQUENCE {
+    modulus           INTEGER,  -- n
+    publicExponent    INTEGER   -- e
+}
+~~~
+
+
+### Deriving the public ECPoint from ECPrivateKey {#sec-ec-pub-from-priv}
+
+Unlike RSA, the ECPrivateKey does not contain sufficient information to simply extract the public key. Note that since a single unique encoding of the traditional public key is required, this specification forbids the optional `publicKey` field.
```
xipki: I do not understand why to forbid the optional publicKey field in this specification. RFC 5915 states: "Though the ASN.1 indicates
publicKey is OPTIONAL, implementations that conform to this
document SHOULD always include the publicKey field."

johngray-dev: Thanks...   I think I messed that text up.  Since the private key isn't in the KEM combiner, the traditional public could be optional in the EC Private Key encoding.  However, if it is optional then it could make incompatibilities.  If an implementation becomes dependent on using it in the private key field, and then it is given one that does have it, will they fail?    Having a single encoding means there won't be two diverging code paths in an implementation waiting to cause an interop problem down the road.

ounsworth: Yeah. I think more simply: the LAMPS WG in Madrid was very clear that we should define a single encoding format for all the things: pub keys, ciphertexts, private keys. So we have been quite clear that anywhere where the existing RFCs allow more than one choice, we have profiled it down to a single choice. Here "Don't include the public key" is the obvious choice.

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -692,10 +670,8 @@ Implicit inputs mapped from <OID>:
 
   tradPK  The traditional public key is required for the KEM
           combiner.
-          The suggested algorithm below extracts the tradPK
-          from sk, however implementations that use a non-standard
-          private key encoding will need to obtain the traditional
-          public key some other way.
+          For discussion of where to get this value, see the
+          discussion below.
```
```suggestion
          For discussion of where to get this value, see the
          Decapsulation Requires the Public Key section.
```

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -692,10 +670,8 @@ Implicit inputs mapped from <OID>:
 
   tradPK  The traditional public key is required for the KEM
           combiner.
-          The suggested algorithm below extracts the tradPK
-          from sk, however implementations that use a non-standard
-          private key encoding will need to obtain the traditional
-          public key some other way.
+          For discussion of where to get this value, see the
+          Decapsulation Requires the Public Key section.
```
wangweij, line 680: `tradPK` cannot be mapped from `<OID>`.

Why not choose the RFC 9180 style to define a `pk(sk)` function?

draft-ietf-lamps-pq-composite-kem.md
wangweij: Left side no longer has `tradPK`.

dconnolly (approved):

draft-ietf-lamps-pq-composite-kem.md
wangweij: But unlike the other 3 implicit inputs, there is no static map to find out `tradPK` from `<OID>`.

RFC 9180 has
> The notation pk(skX), depending on its use and the KEM and its implementation, is either the computation of the public key using the private key, or just syntax expressing the retrieval of the public key, assuming it is stored along with the private key object.

I think this is exactly what is needed here. Not necessarily computed from the private key, it can also be stored along with it.

ounsworth: (I deleted my previous comment).

Thanks. Adjusted.

We added a whole appendix talking about how to extract TradPK from Trad SK.

Issue #239, opened 2025-11-03
Address Peter C's WGLC comments


danvangeest:
Peter C's WGLC comments: https://mailarchive.ietf.org/arch/msg/spasm/erQjg2YMk5hjpdZoJXiB0ophnUU/

feventura:
Summarizing Peter C non-minor comments in suggested actions:

- [ ] Need proof.
  - [ ] Suggest delaying WGLC if the proof is close to be done, so it can be reviewed. If is not add the accurate situation to Security Considerations and not point to a yet-not-existing proof.
  - [ ] Does not prove that the ML-KEM ciphertext can be omitted in general. (Section 10.2.2)
  - [x] The X-Wing proof does not assume DH is IND-CCA. (Section 10.2.1)
- [x] No mention of input validation
  -  This is important to FIPS certification
- [x] FIPS 203 considers cryptographic input validation to be external to ML-KEM key generation, encapsulation and decapsulation.  It may not be safe to assume that, for example, ML-KEM decapsulation will catch a malformed private key. (Section 5)
- [x] Labels should be distinct.
- [x] Comment about encaps side channel attack timing attack that would reveal which component algorithm fail does not help the attacker recover composite shared secret from a successful one because each call to the composite encapsulation algorithm uses fresh randomness, so they are mutually independent. (Section 4.2)
- [x] All inputs are fixed, except when they are not. Suggestion to delete the fixed length text unless its needed for security (Section 4.3)
- [x] Comment about discovering which component fail by a timing attack. (Section 4.3)
- [x] X-wing has LEAK-C2PRI not MAL-C2PRI ("or wrose" text). (Section 10.2)

feventura:
> Section 4.3: “In order to properly achieve its security properties, then KEM combiner requires that all inputs are fixed-length […] all inputs should be fixed-length in non-error scenarios except for minor variations introduced by encoding.”  There is a clear disconnect here: all inputs must be fixed length, but some might not be.  Unless fixed-length inputs are genuinely necessary for security, I would delete the first sentence.

Later in the text its clarified that these minor variations are not a issue for the security properties:

>In the cases where there are minor variations introduced by encoding, those encodings already have a fixed-length prefix followed by length-encoded data, so the requirements for the KEM combiner security properties hold (namely that the input is injective).

[...]

PR #245, opened 2025-11-06
Removed text in accordance with WGLC comments

feventura:
Removed "timing-invariant way to prevent side-channel attackers" text.


johngray-dev (approved):
I agree with these changes.  Either we explicitly state there is a timing attack, or just not include this text.    Also, I agree with the algorithm is completely broken that is bad enough to consider for the case

ounsworth:
Relates to, but does not completely close #239

ounsworth (approved):

PR #260, opened 2026-01-15
Adds explicit reference to ML-KEM seed (d, z).

### LINKED ISSUE #259: Clarify relationship to ML-KEM key generation interface
In s3.1 of the -12 draft:

```
     1. Generate component keys

       mlkemSeed = Random(64)
       (mlkemPK, mlkemSK) = ML-KEM.KeyGen(mlkemSeed)
       (tradPK, tradSK) = Trad.KeyGen()
```

In s3.1.1:

```
   Implementations MAY modify this process to additionally output the
   expanded mlkemSK or to make use of ML-KEM.KeyGen_internal(mlkemSeed)
   as needed to expand the ML-KEM seed into an expanded key prior to
   performing a signing operation.
```

In s3.3:

```
     1. Separate the private keys and ciphertexts

         (mlkemSeed, tradSK) = DeserializePrivateKey(sk)
         (_, mlkemSK) = ML-KEM.KeyGen(mlkemSeed)
         (mlkemCT, tradCT) = DeserializeCiphertext(ct)
```

In s10.1:

```
   from the one under which it was certified.  For example, the KeyGen
   defined in Section 3.1 invokes ML-KEM.KeyGen(seed) which might not be
   available in a cryptographic module running in FIPS-mode, but
   Section 3.1 is only a suggested implementation and the composite
   KeyGen MAY be implemented using a different available interface for
   ML-KEM.KeyGen.
```

ML-KEM.KeyGen exists in FIPS-203, but it does not take any parameters. Probably what is meant is ML-KEM.KeyGen_internal, which takes _two_ parameters, d and z. Presumably mlkemSeed is meant to be the concatenation of these two values, in some order.

I think the hybrid spec should use ML-KEM.KeyGen_internal intead of ML-KEM.KeyGen, and make explicit that mlkemSeed = d || z.

janklaussner:
The text now refers to the seed definition (d, z) according to FIPS.203 and its usage in the internal KeyGen function.

Closes #259

### DIFF
--- a/draft-ietf-lamps-pq-composite-kem.md
+++ b/draft-ietf-lamps-pq-composite-kem.md
@@ -523,6 +523,8 @@ non-composite keys, or between multiple composite keys. This means that an invoc
 
 To generate a new keypair for composite schemes, the `KeyGen() -> (pk, sk)` function is used. The KeyGen() function calls the two key generation functions of the component algorithms independently. Multi-threaded, multi-process, or multi-module applications might choose to execute the key generation functions in parallel for better key generation performance or architectural modularity.
 
+To generate an ML-KEM key pair this specification uses the function `ML-KEM.KeyGen_internal(d, z)`. According to [FIPS.203] d and z are two random 32 Byte values. This document combines both values in mlkemSeed by concatenating them so that `mlkemSeed = d || z`.
+
 The following describes how to instantiate a `KeyGen()` function for a given composite algorithm represented by `<OID>`.
 
 ~~~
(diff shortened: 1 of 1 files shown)


optnfast:
✅✅✅✅

draft-ietf-lamps-pq-composite-kem.md
```diff
@@ -548,7 +550,7 @@ Key Generation Process:
   1. Generate component keys
 
     mlkemSeed = Random(64)
-    (mlkemPK, mlkemSK) = ML-KEM.KeyGen(mlkemSeed)
+    (mlkemPK, mlkemSK) = ML-KEM.KeyGen_internal(mlkemSeed[:32], mlkemSeed[32:])
```
johngray-dev, line 553: Couldn't it just use the ML-KEM.KeyGen() method without requiring the random input as well? That would work better for black-box implementations.

johngray-dev: I guess the above is just a suggestion. A black box could still just to ML-KEM.KeyGen() and let the crypto module create the random data.

johngray-dev (approved):

PR #267, opened 2026-06-02
Fix ML-KEM automated verifier

danvangeest:


johngray-dev (approved):
Thanks for the fix Dan.  It looks good to me.

